PR #2, opened 2019-02-23
Setup CircleCI

joneshf-cn:
##### What it does

Sets up CircleCI!


joneshf-cn:
For audit, got a quick review from Forest in slack.

PR #3, opened 2019-02-23
Setup test suite

joneshf-cn:
##### What it does

Pulls the test JSON values out to the test suite.

##### Still to do

* [x] Move the JSON from `Data.Expr`.

### DIFF
--- a/src/Data/Form.purs
+++ b/src/Data/Form.purs
@@ -7,7 +7,7 @@ import Data.Either (Either(..))
 import Data.Generic.Rep (class Generic)
 import Data.Generic.Rep.Show (genericShow)
 import Data.Maybe (Maybe(..))
-import Lynx.Data.Expr (Expr, ExprF(..), mkExpr)
+import Lynx.Data.Expr (Expr, val_)
 import Type.Row (type (+))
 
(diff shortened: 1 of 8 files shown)


src/Data/Form.purs
```diff
@@ -185,69 +185,73 @@ instance decodeInputSource :: DecodeJson InputSource where
...
-"""
 
-toggleInput :: String
-toggleInput = """
-  { "type": "Toggle"
-  , "value": { "value": null, "source": null }
-  , "default": """ <> val false "Boolean" <> """
+firstName :: Field Expr
+firstName =
+  { name: mkExpr (Val identity "First Name")
```
davezuch: I know this test data isn't meant to stick around, but while we're manually constructing these, it might be worthwhile adding a helper function so we can just do something like
```purescript
  name: val_ "First Name"
```

davezuch (approved):

src/Data/Form.purs
joneshf-cn: Good idea! I went ahead and made a helper for all the cases since we're most likely going to be writing them once we build momentum.

davezuch (approved):
Nice, gonna merge this

PR #4, opened 2019-02-27
Make `Expr` untyped

joneshf-cn:
##### What it does

Tries an approach to encoding expressions as untyped. The motivation behind this change is that the well-typed expressions were giving us issues when attempting to evaluate them. If we have no real types, we can always attempt to evaluate something. It ends up that we have to deal with failure in the cases where we've constructed an ill-typed expression, though.

##### Review asks

Is this approach something we want to continue pursuing? I'm more than happy to close this PR in exchange for something else, but I'm not sure what that something else might be.

There are a few open questions that I couldn't figure out on my own. I left them as review notes in this PR.

### DIFF
--- a/src/Data/Expr.purs
+++ b/src/Data/Expr.purs
@@ -31,10 +31,17 @@ reflectType (Boolean _) = "Boolean"
 reflectType (Int _) = "Int"
 reflectType (String _) = "String"
 
+print :: ExprType -> ExprType
+print = case _ of
+  Boolean x -> String (show x)
+  Int x -> String (show x)
+  String x -> String x
+
 data Expr
   = Val ExprType
   | If Expr Expr Expr
   | Equal Expr Expr
+  | Print Expr
   | Lookup Key
 
 derive instance genericExpr :: Generic Expr _
@@ -67,6 +74,7 @@ data ExprA
   = ValA ExprType
   | IfA ExprA ExprA ExprA String String
   | EqualA ExprA ExprA String String
+  | PrintA ExprA String String
   | LookupA Key String
 
 instance encodeJsonExprA :: EncodeJson ExprA where
@@ -77,6 +85,8 @@ instance encodeJsonExprA :: EncodeJson ExprA where
         "params" := [ x, y, z ] ~> encodeHelper "If" i o
       EqualA x y i o ->
         "params" := [ x, y ] ~> encodeHelper "Equal" i o
+      PrintA x i o ->
+        "params" := [ x ] ~> encodeHelper "Print" i o
       LookupA x o -> "param" := x ~> encodeHelper "Lookup" "Void" o
 
 toExprA :: Expr -> ExprA
@@ -84,6 +94,7 @@ toExprA = case _ of
   Val x -> ValA x
   If x y z -> IfA (toExprA x) (toExprA y) (toExprA z) (reflectIn x) (reflectOut y)
   Equal x y -> EqualA (toExprA x) (toExprA y) (reflectIn x) (reflectOut x)
+  Print x -> PrintA (toExprA x) (reflectIn x) "String"
   Lookup x -> LookupA x "???"
   where
   reflectIn :: Expr -> String
@@ -91,12 +102,14 @@ toExprA = case _ of
     Val _ -> "Void"
     If x _ _ -> reflectIn x
     Equal x _ -> reflectIn x
+    Print x -> reflectIn x
     Lookup _ -> "Void"
   reflectOut :: Expr -> String
   reflectOut = case _ of
     Val x -> reflectType x
     If _ x _ -> reflectOut x
     Equal _ _ -> "Boolean"
+    Print x -> "String"
     Lookup _ -> "???"
 
 encodeHelper :: String -> String -> String -> Json
@@ -123,6 +136,9 @@ evalExpr get = case _ of
     Right left, Right right -> Left (EqualMismatch { left, right })
     Left x, _ -> Left x
     _, Left x -> Left x
+  Print x' -> case evalExpr get x' of
+    Right x -> Right (print x)
+    Left x -> Left x
   Lookup x -> maybe (Left $ MissingKey x) Right (get x)
 
 boolean_ :: Boolean -> Expr
--- a/src/Data/Expr.purs
+++ b/src/Data/Expr.purs
@@ -2,6 +2,7 @@ module Lynx.Data.Expr where
 
 import Prelude
 
+import Data.Argonaut (class DecodeJson, class EncodeJson, Json, decodeJson, encodeJson, jsonEmptyObject, (.:), (:=), (~>))
 import Data.Either (Either(..))
 import Data.Generic.Rep (class Generic)
 import Data.Generic.Rep.Show (genericShow)
@@ -19,6 +20,17 @@ derive instance genericExprType :: Generic ExprType _
 instance showExprType :: Show ExprType where
   show = genericShow
 
+instance encodeJsonExprType :: EncodeJson ExprType where
+  encodeJson = case _ of
+    Boolean x -> encodeJson x
+    Int x -> encodeJson x
+    String x -> encodeJson x
+
+reflectType :: ExprType -> String
+reflectType (Boolean _) = "Boolean"
+reflectType (Int _) = "Int"
+reflectType (String _) = "String"
+
 data Expr
   = Val ExprType
   | If Expr Expr Expr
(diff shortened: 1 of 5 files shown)


src/Data/Expr.purs
```diff
@@ -4,259 +4,157 @@ import Prelude
...
+        "params" := [ x ] ~> encodeHelper "Print" i o
+      LookupA x o -> "param" := x ~> encodeHelper "Lookup" "Void" o
+
+toExprA :: Expr -> ExprA
+toExprA = case _ of
+  Val x -> ValA x
+  If x y z -> IfA (toExprA x) (toExprA y) (toExprA z) (reflectIn x) (reflectOut y)
+  Equal x y -> EqualA (toExprA x) (toExprA y) (reflectIn x) (reflectOut x)
+  Print x -> PrintA (toExprA x) (reflectIn x) "String"
+  Lookup x -> LookupA x "???"
```
joneshf-cn: ❓ What do we do here? Does `LookupA` have an output type? Is it `Unit`? Is it something else?

src/Data/Expr.purs
```diff
@@ -4,259 +4,157 @@ import Prelude
...
-class (Show a, Eq a) <= Expressible a where
-  toExprType :: a -> ExprType
-  reflectProxy :: Proxy a -> String
-  print :: a -> String
-
-instance exprBoolean :: Expressible Boolean where
-  toExprType = Boolean
-  reflectProxy _ = "Boolean"
-  print = show
+print :: ExprType -> ExprType
```
joneshf-cn: I think this is typed not right, it was thrown together when adding `Print` back in the mix, but it looks like we probably want the type to be `ExprType -> String` and we can do the wrapping where it's used.

src/Data/Expr.purs
```diff
@@ -4,259 +4,157 @@ import Prelude
...
-decodeEqualF xs = Left "Expected 2 params"
+data EvalError
+  = MissingKey Key
+  | IfCondition ExprType
+  | EqualMismatch { left :: ExprType, right :: ExprType }
+
+evalExpr :: (Key -> Maybe ExprType) -> Expr -> Either EvalError ExprType
+evalExpr get = case _ of
+  Val x -> pure x
+  If x' y z -> case evalExpr get x' of
```
joneshf-cn: I think the explicit casing on the `Either _ _` can be replaced with binds. Maybe something like:

```PureScript
If x' y z -> do
  x <- evalExpr get x'
  case x of
    Boolean false -> evalExpr get z
    Boolean true -> evalExpr get y
    _ -> Left (IfCondition x)
```
or
```PureScript
If x' y z -> evalExpr get x' >>= case _ of
  Boolean false -> evalExpr get z
  Boolean true -> evalExpr get y
  x -> Left (IfCondition x)
```

Not a huge deal on this one, but it might clean up the `Equal` case a bit.

src/Data/Form.purs
```diff
@@ -138,23 +58,18 @@ type InputState t r =
   )
 
 data Input f
-  = Text (Record (SharedRows f String + RequiredRows f + StringRows f ()))
-  | Toggle (Record (SharedRows f Boolean ()))
+  = Text (Record (SharedRows f + RequiredRows f + StringRows f ()))
+  | Toggle (Record (SharedRows f ()))
 
-derive instance genericInput :: Generic (Input f) _
+derive instance genericInput :: Generic (Input Expr) _
```
joneshf-cn: I changed this on accident. Maybe we can remove the change so the diff is a little cleaner?

```suggestion
derive instance genericInput :: Generic (Input f) _
```

src/Data/Form.purs
```diff
@@ -138,23 +58,18 @@ type InputState t r =
...
-  -- hmap f (Text r) = Text $ hmap f r
-  -- hmap f (Toggle r) = Toggle $ hmap f r
 
 instance encodeInput :: EncodeJson (Input Expr) where
   encodeJson = case _ of
     Text r -> "type" := "Text" ~> encodeJson r
     Toggle r -> "type" := "Toggle" ~> encodeJson r
 
-instance decodeInput :: DecodeJson (Input Expr) where
+instance decodeJsonInput :: DecodeJson (Input Expr) where
```
joneshf-cn: Similar accidental change.

```suggestion
instance decodeInput :: DecodeJson (Input Expr) where
```

src/Data/Form.purs
```diff
@@ -243,15 +158,19 @@ lastName =
...
-    { default: Just (val_ false)
+    { default: Just (boolean_ true)
     , value:
       { source: Nothing
       , value: Nothing
       }
     }
   }
+  where
+  description = if_ (equal_ (boolean_ true) (lookup_ "active"))
```
joneshf-cn: I wonder if we can drop the `equal_` stuff? It should be semantically the same as `if_ (lookup_ "active")`, right?
```suggestion
  description = if_ (lookup_ "active")
```

src/Page/Form.purs
```diff
@@ -19,12 +24,14 @@ import Ocelot.Block.Toggle (toggle) as Toggle
 import Ocelot.HTML.Properties (css)
 
 type State =
-  { form :: Either String (Page Expr)
+  { form :: Either String { expr :: Page Expr, evaled :: Page ExprType }
```
joneshf-cn: I'm not sold on the naming here. Happy to change it to whatever if we've got better names. Also fine to keep it as is, if we don't have anything better.

src/Page/Form.purs
```diff
@@ -47,18 +54,105 @@ component =
...
-  eval (EvalForm form a) = a <$ do
-    H.modify_ _ { form = pure form }
-    pure unit
+  eval (EvalForm expr a) = a <$ do
+    let evaled' = evalPage get expr
+        get key = Data.Map.lookup key values
+        values = keysPage expr
+    H.modify_ _ { values = values }
+    case evaled' of
+      Left _ -> pure unit
```
joneshf-cn: ❓ This ignores any errors that we had when we're evaluating the form. What should we do in this case? Do we want to throw the message up in a toast or something? Do we log errors somewhere we can later take a look at?

src/Page/Form.purs
```diff
@@ -47,18 +54,105 @@ component =
...
+  keysPage :: Page Expr -> Map Key ExprType
+  keysPage page = foldMap keysSection page.contents
+
+  keysSection :: Section Expr -> Map Key ExprType
+  keysSection section = foldMap keysField section.contents
+
+  keysField :: Field Expr -> Map Key ExprType
+  keysField field = case field.input of
+    Text { value: { value: Just expr } }
+      | Right value <- evalExpr (const Nothing) expr ->
```
joneshf-cn, line 137: ❓ These `const Nothing`s are actually limiting the amount of depth we're able to have for lookups. If the expression we're looking up is a lookup itself, it won't find the key, even if it exists. Maybe there's a different way to do this? Maybe we should ditch the idea of a map in favor of the pub/sub stuff?

joneshf-cn (commented):
Giving myself a once-over for to-dos so I don't forget them.

I'll address what I can and leave the unaddressed stuff for review.

src/Data/Expr.purs
davezuch: I think this should be the same type of whatever value you're looking up. Which should be known at encode time, since each input type has only one valid type. Also, there are type constraints on where in an expression it's used. 

Some clarifying examples:
```purescript
{ name: string_ "Name"
, visibility: lookup_ "active"
, description: if_ (lookup_ "active") (string_ "you need a name") (lookup_ "foo")
}
```
So above,  `visibility` needs to be a Boolean, so only boolean lookups are valid here, so this could only have been encoded if `active` is a Boolean input. Similarly, in `description`, the `active` lookup is valid because an `if_` condition has to be a Boolean expression.

src/Page/Form.purs
davezuch: This is better than anything I came up with when I was trying lol

src/Page/Form.purs
davezuch: At this point an error is fatal, so I think we could replace the whole `case` with
```purescript
H.modify_ _ { form = { expr, evaled: _ } <$> evaled }
```
And update the type of our state so the `Left` is an actual error type

src/Page/Form.purs
davezuch: Hmm, I'm not sure our lookups need to be recursive. I mean, that would be pretty fancy, but the results of our lookups will be the value of a form field, and I'm not really imagining allowing users to build up expressions in the generated form. Although...that would be pretty mind blowing 🤔

src/Page/Form.purs
```diff
@@ -47,18 +54,105 @@ component =
...
+    let evaled' = evalPage get expr
+        get key = Data.Map.lookup key values
+        values = keysPage expr
+    H.modify_ _ { values = values }
+    case evaled' of
+      Left _ -> pure unit
+      Right evaled -> H.modify_ _ { form = pure { expr, evaled } }
+
+  eval (UpdateKey key val a) = do
+    form <- H.gets _.form
```
davezuch: This is a tiny, mostly cosmetic thing, but I like to write these as
```purescript
{ form } <- H.get
```
I think it's a little cleaner, and allows you to match multiple values more efficiently if you needed to do something like
```purescript
{ form, values } <- H.get
```

src/Data/Expr.purs
joneshf-cn: Hmm, it seems like we'd have to have the `Page Expr` then to do that, right? Or at least some `Map Key String`/`Key -> String` or whatever. The `Key`s only exist in a `Page Expr`, so maybe we can't have an instance for `EncodeJson Expr`. It would have to be `EncodeJson (Page Expr)` and friends, so they could traverse with the type context.

That doesn't seem too bad, since `ExprA`s reason for existence seems to be to facilitate encoding.

src/Page/Form.purs
joneshf-cn: Oh right, that makes sense! The thing we're looking up is an `ExprType` anyway, not an `Expr`. Sorry about that.

src/Data/Expr.purs
davezuch: I see what you mean, with `ExprF i o` we never lose our output type but if we drop our types then we have no way of recovering that information without additional context.

One trick we could try would be to have a failure case for our lookup, something like
```purescript
  | Lookup Key Expr
```
That way we can kind of cheat and look at the output type of the failure case Expr.

joneshf-cn: That seems like it'll also remove one more error case to deal with. Lemme give it a shot.

joneshf-cn: That seemed to work out! Great suggestion!

joneshf-cn:
I think I addressed all the comments. Lemme know if there's more we should do here. If not, I'll merge this after CI passes.

davezuch (approved):
I'm happy with this. Honestly this change to dynamic typing seems quite promising, at the least it's easier to work with.

PR #5, opened 2019-03-01
Switch over to a `Makefile`

joneshf-cn:
##### What it does

As discussed on slack, we're going to try using a `Makefile` for building PS stuff. This is mostly a direct port of the npm scripts. It does set up dependencies automatically though. I didn't want to spend too much time on this, so it's still pretty naïve. It could be made a bit better by using `purs` directly instead of going through `pulp` still.

More than happy to explain more motivation if it's not clear from the commit/slack conversation.

##### Review asks

Are we cool with moving to a `Makefile`?

Requesting a review from Dave (since you're working a bunch in this repo) and Forest (since we're in this escapade together).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,76 @@
+# There are a couple of conventions we use so make works a little better.
+#
+# We sometimes want to build an entire directory of files based on one file.
+# We do this for dependencies. E.g.: package.json -> node_modules.
+# For these cases, we track an empty `.stamp` file in the directory.
+# This allows us to keep up with make's dependency model.
+#
+# We also want some place to store all the excess build artifacts.
+# This might be test outputs, or it could be some intermediate artifacts.
+# For this, we use the `$(BUILD)` directory.
+# Assuming the different tools allow us to put their artifacts in here,
+# we can clean up builds really easily: delete this directory.
+#
+# We use some make syntax that might be unfamiliar, a quick refresher:
+#
+# <target>: <prerequisites> | <order-only-prerequisites>
+# 	<command>
+#
+# `<order-only-prerequisites>` are similar to normal `<prerequisites>`
+# but they don't cause a target to be rebuilt if they're out of date.
+# This is mostly useful for creating directories and whatnot.
+#
+# See: https://www.gnu.org/software/make/manual/make.html#Prerequisite-Types
+#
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -0,0 +1,76 @@
...
+	  node_modules
+
+dist/main.js: $(OUTPUT)/Main/index.js
+	npx pulp build --to $@
+
+test: dist/main.js $(BUILD)/test.out
+
+.PHONY: watch
+watch: $(BOWER_COMPONENTS) $(NODE_MODULES)
+	npx pulp --watch build --to dist/main.js
```
davezuch: I've always had problems with this watch, sometimes it rebuilds twice (on any of our EC2's at least, didn't experience locally), sometimes doesn't build at all. I saw some people in Slack (you may have been one of them) suggest `watch-exec` instead, and that's worked beautifully. It is an extra dependency, but might be worth using instead

davezuch (approved):
Sorry it took me a while to review, I wasn't very familiar with `make`, so had to do a little reading when I didn't get something. This looks dope though 👍

foresttoney (approved):
This all looks good to me

Makefile
foresttoney: I have also seen this behavior - I didn't know there was something I could install to make my life better.

joneshf-cn: It looks simple enough to add, so I'm down to give it a shot!

joneshf-cn:
> I wasn't very familiar with `make`, so had to do a little reading when I didn't get something.

I'm more than happy to add whatever caught you up to the `Makefile` as documentation. It's likely someone else will get bit by it as well. What stuff should we add?

davezuch (approved):

davezuch:
> I'm more than happy to add whatever caught you up to the Makefile as documentation. It's likely someone else will get bit by it as well. What stuff should we add?

The comments and links in your Makefile were enough for me. Just taking the time to read the more detailed explanations and parse the examples is what did it. If anything I might explain or link to documentation for a few more things like the `.DEFAULT_GOAL` and `.PHONY` keywords, but that's all available when you follow any of the links

joneshf-cn:
Expanded on the docs. Lemme know if there's more we should follow up with.

PR #6, opened 2019-03-04
Simplify value

joneshf-cn:
##### What it does

Through discussions, we realized that the value had more cases than we wanted to handle. We changed the types here to reflect what we want to handle.

Also sets up some round-trip tests for the JSON decoding/encoding. Turns out we weren't doing any decoding of prints.

### DIFF
--- a/src/Data/Form.purs
+++ b/src/Data/Form.purs
@@ -2,11 +2,13 @@ module Lynx.Data.Form where
 
 import Prelude
 
-import Data.Argonaut (class DecodeJson, class EncodeJson, decodeJson, encodeJson, fromString, (.:), (:=), (~>))
+import Data.Argonaut (class DecodeJson, class EncodeJson, decodeJson, encodeJson, jsonEmptyObject, (.:), (:=), (~>))
 import Data.Either (Either(..))
+import Data.Foldable (class Foldable, foldlDefault, foldrDefault)
 import Data.Generic.Rep (class Generic)
 import Data.Generic.Rep.Show (genericShow)
 import Data.Maybe (Maybe(..))
+import Data.Traversable (class Traversable, sequenceDefault)
 import Lynx.Data.Expr (Expr, boolean_, if_, lookup_, string_)
 import Type.Row (type (+))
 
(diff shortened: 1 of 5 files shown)


src/Data/Form.purs
```diff
@@ -35,7 +39,7 @@ type Field f = Record (FieldRows f ())
 
 type SharedRows f r =
   ( default :: Maybe f
-  , value :: Record (InputState f ())
+  , value :: Maybe (InputSource f)
```
davezuch: I was thinking a little more about how we're modeling this. Ultimately we have three possible value here
```purescript
Nothing | Just (UserInput a) | Just (Invalid a)
```
I'm just slightly bothered by having an invalid state in right-hand side (`Just` case). Seems like we could get the same results by rearranging it like
```purescript
Left NotSet | Left (Invalid a) | Right a
```
The benefit here is better separation between valid and invalid values. We can use `map` to only map over valid values, or `isLeft` to know if our value is invalid and we should resort to our `default`.

I wonder what adding validation support will do here, if anything. Not sure yet if it should be coupled to this value or not, but I can imagine expanding the sum type we use on the left for other possible error states.

davezuch (approved):

src/Data/Form.purs
joneshf-cn: I'm with it. Maybe instead of bringing in the connotations of `Maybe _`/`Either _ _`, we enumerate the cases at our domain level?
```PureScript
data Value a
  = NotSet
  | Invalid a
  | UserInput a
```

My reasoning is, `NotSet` should only be an error if the field is required and there's no default value. If the field isn't required or there's a default value, it seems fine to have the value be `NotSet`.

What do you think?

davezuch: Hmm, yeah you're right `NotSet` isn't necessarily an error, but there's still a distinction between having a value and not. If we drop the `Either` we lose the bifunctor machinery, but maybe that's okay. We can always add it back if we decide it'd be more useful or a better representation, refactoring is cheap 😉

joneshf-cn: K. I'll switch it over and merge afterward.

davezuch (approved):
Solid!

PR #7, opened 2019-03-06
Render default values

davezuch:
## What it does

 Adds rendering for default values when a user value isn't set. Also adds a number of helper functions for dealing with values.

Also introduces templates for new PR's and Issues. Perhaps they shouldn't be coupled to the same PR...


src/Page/Form.purs
```diff
@@ -184,8 +184,8 @@ component =
   renderField :: Field ExprType -> H.ComponentHTML Query
   renderField (field) =
     FormField.field_
-      { label: HH.text $ show field.name
-      , helpText: Just $ show field.description
+      { label: HH.text $ print field.name
```
joneshf-cn, line 187: 😻 Thanks for doing this. It's a small change, but I feel like it's enough to make things look nicer.

joneshf-cn (approved):
Looks great! Thanks!

PR #8, opened 2019-03-06
PS-2958 Implement dropdown

joneshf-cn:
##### What it does

This gives us support for dropdowns. It seemed too easy so I probably missed some stuff. The long and short is that there's a bunch of questions 😅.

##### Review asks
* <strike>The  `VisibilityChanged _` message, are we supposed to be setting the visibility in the form based on that?</strike> Threaded through the value down to the field. If that's not right, let's back out those changes.
* <strike>I was a little confused about the `o` we set for `Dropdown.Message`/`Dropdown.Query`. In other places we use the `Query` we're defining, like:
    ```PureScript
    data Query a
      = Foo a
      | Dropdown (Dropdown.Message Query Int) a
    ```
    When should we do that? Should we always do that? What's the intent with the `Emit _` case?</strike>. Read through the [halogen-select guide][]. It's all clear to me now.
* What is the `valueRepresentation` supposed to be? If we've got an `Array { name :: String, value :: ExprType }`, can we render them directly and drop the idea of `valueRepresentation`? If we need it for something, it seems like it should be a function from `Expr -> String`  or `ExprType -> String` or something of that nature. How would we encode that as JSON?

[halogen-select guide]: https://citizennet.github.io/purescript-halogen-select


davezuch:
> What is the `valueRepresentation` supposed to be? If we've got an `Array { name :: String, value :: ExprType }`, can we render them directly and drop the idea of valueRepresentation? If we need it for something, it seems like it should be a function from `Expr -> String or ExprType -> String` or something of that nature. How would we encode that as JSON?

It might not be necessary for MVP, or possibly even dropdowns. Where this really becomes necessary is for typeaheads where we don't know the shape of our data. `valueRepresentation` allows the user to specify how we render foreign data.

For dropdowns with values that are always in the shape of `{ name :: String, value :: ExprType }` we can probably get away with rendering just the name. However, in some cases it might be useful to allow the user to say define rendering like:
```purescript
option = { name: "John Doe", value: Int 1324 }
render = (concat_
  (lookup_ "name")
  (concat_
    (string_ " (")
    (concat_
      (lookup_ "value")
      (string_ ")")
    )
  )
)
output = "John Doe (1324)"
```
The thing to note about this expression is it's scoped to the current option, so `lookup_` looks up the key on the option instead of on the form.

Another note is that eventually we'll probably want to support more advanced rendering, probably Markdown, if not some sort of template HTML. I think Markdown would be enough though.

src/Data/Form.purs
```diff
@@ -55,8 +55,17 @@ type StringRows f r =
   | r
   )
 
+type DropdownRows f r =
+  ( index :: Int
+  , options :: Array { name :: String, value :: f }
```
davezuch: `options` will actually need to be an `Expr`. Which means we'll probably need to add `Array` as a supported `ExprType`.
 
The motivation is so that we can have different options for a dropdown given different form states. E.g.:

```purescript
options:
  (if_ (equal_ (lookup_ "meal") (string_ "Breakfast"))
    (array_
      [ pair_ "Pancakes" (string_ "Pancakes")
      , pair_ "Eggs & Toast" (string_ "EggsAndToast")
      ]
    )
    (array_
      [ pair_ "Salad" (string_ "Salad")
      , pair_ "Sandwich" (string_ "Sandwich")
      ]
    )
  )
```

I dunno that `Pair` is the best constructor for a key/value pair but that was the best I could come up with.

Also, we can get away with nested if/else chains here, but it might be more ergonomic to extend our algebra with `case`s, so we can extend the above as:

```purescript
options:
  (case_ (lookup_ "meal")
    [ (string_ "Breakfast") (array_ ...)
    , (string_ "Lunch") (array_ ...)
    , (string_ "Snack") (array_ ...)
    ]
    (array_ ...) -- otherwise/failure case
  )
```

davezuch: If you'd like to get a better understanding of the need for dynamic options, try loading up a new campaign form in Wildcat, then changing the `Objective`. You'll notice the available options (as well as the default values) for `Optimization Goal` and `Billing Event` will change (and for some values certain fields like `Application` might show or hide).

https://citizennet.com/manager/#/groups/facebook/159675/campaigns/new/details

src/Data/Form.purs
```diff
@@ -127,6 +138,126 @@ instance decodeInputSource :: (DecodeJson a) => DecodeJson (InputSource a) where
 instance arbitraryInputSource :: (Arbitrary a) => Arbitrary (InputSource a) where
   arbitrary = genericArbitrary
 
+userInput :: forall a. InputSource a -> Maybe a
+userInput = case _ of
+  UserInput x -> Just x
+  Invalid _ -> Nothing
+  NotSet -> Nothing
```
davezuch: Hmm, now that we have this helper we can probably simplify my `getValue` helper to: 
```purescript
getValue { value, default } = userInput value <|> defualt
```

davezuch: Might also be able to shorten this to
```purescript
userInput (UserInput x) = Just x
userInput _ = Nothing
```
Since judging from the function name everything but `UserInput x` should fail.

src/Data/Form.purs
```diff
@@ -127,6 +138,126 @@ instance decodeInputSource :: (DecodeJson a) => DecodeJson (InputSource a) where
...
+    Text { default: Just expr }
+      | Right value <- evalExpr (const Nothing) expr ->
+        Data.Map.singleton field.key value
+    Toggle { value: UserInput expr  }
+      | Right value <- evalExpr (const Nothing) expr ->
+        Data.Map.singleton field.key value
+    Toggle { default: Just expr }
+      | Right value <- evalExpr (const Nothing) expr ->
+        Data.Map.singleton field.key value
+    _ -> mempty
```
davezuch: Don't we need a `Dropdown` case here?

davezuch: Seems we could cut these in half by using the `getValue` helper

src/Page/Form.purs
```diff
@@ -58,91 +76,39 @@ component =
...
-    Toggle { default: Just expr }
-      | Right value <- evalExpr (const Nothing) expr ->
-        Data.Map.singleton field.key value
-    _ -> mempty
-
-  render :: State -> H.ComponentHTML Query
+  eval (DropdownQuery key message a) = case message of
+    Dropdown.Emit x -> a <$ eval x
+    Dropdown.Selected val -> eval (UpdateKey key val a)
+    Dropdown.VisibilityChanged visibility' -> do
```
davezuch: I think we want to ignore this query. Visibility on a dropdown is different from visibility on a field. This query let's us know when the dropdown container's visibility has been toggled on or off. So if a user clicks on the dropdown, the container's visibility will toggle `On` and the options will be shown, then if an option is selected or the user clicks elsewhere, the visibility will toggle `Off` and the container will hide.

Separate from that is our notion of `visibility` on a field, which is an expression to determine whether the field itself should be rendered or not.

davezuch:
Also, sorry, I merged my other PR and now there are conflicts 😔

joneshf-cn:
> It might not be necessary for MVP, or possibly even dropdowns. Where this really becomes necessary is for typeaheads where we don't know the shape of our data. `valueRepresentation` allows the user to specify how we render foreign data.

Ah! That makes sense. Yeah seems like for MVP, we can get away with not dealing with it.

> For dropdowns with values that are always in the shape of `{ name :: String, value :: ExprType }` we can probably get away with rendering just the name. However, in some cases it might be useful to allow the user to say define rendering like:
> 
> ```purescript
> option = { name: "John Doe", value: Int 1324 }
> render = (concat_
>   (lookup_ "name")
>   (concat_
>     (string_ " (")
>     (concat_
>       (lookup_ "value")
>       (string_ ")")
>     )
>   )
> )
> output = "John Doe (1324)"
> ```
> 
> The thing to note about this expression is it's scoped to the current option, so `lookup_` looks up the key on the option instead of on the form.

For clarity, if the options can be expressions, are their lookups also scoped to themselves or are they scoped to the entire form?

Also, if options are expressions, how would we know how to lookup something in them? Do we need records in our expressions?

> Also, sorry, I merged my other PR and now there are conflicts 😔

No worries!

davezuch:
> For clarity, if the options can be expressions, are their lookups also scoped to themselves or are they scoped to the entire form?

So far the only lookup that is locally scoped is for `valueRepresentation`, the expressions for `options` would be scoped to the entire form like other expressions.

> Also, if options are expressions, how would we know how to lookup something in them? Do we need records in our expressions?

Hmm, not sure I follow, but we would need to ensure from the form builder that their expressions are sane, and will ultimately evaluate to an array of key/value pairs. Not sure if I answered your question?

src/Data/Form.purs
joneshf-cn: > Which means we'll probably need to add `Array` as a supported `ExprType`.

If we make arrays `ExprType`s, it doesn't seem like they would be able to have `Expr`s for values. Is that right?

> it might be more ergonomic to extend our algebra with `case`s

Can this be sugar outside the core of the expressions? It seems like we can convert cases to ifs:

```PureScript
case_ :: Expr -> NonEmptyArray (Tuple Expr Expr) -> Expr -> Expr
case_ cond cases def = if_ (equal_ cond ?firstCondition) ?firstResult ?iterateTheRestOfTheCases
```

joneshf-cn: Oh, I guess it could be:
```PureScript
data ExprType
  = Boolean Boolean
  | Int Int
  | String String
  | Array (Array Expr)
```
Nevermind.

src/Data/Form.purs
joneshf-cn: We sure do need that case!

src/Data/Form.purs
joneshf-cn: The [bug][] you pointed out below is the main reason I try to write all the cases explicitly when I remember. I didn't in that case, and luckily you caught it. But, this is work the compiler should be doing for us. I'm of two minds with this: I don't enjoy writing the cases explicitly, but I really dislike having to remember to check cases. If you're alright with it, can we keep this one explicit so we're less likely to run into another one of these [bugs][bug]?

[bug]: https://github.com/citizennet/purescript-lynx/pull/8#discussion_r263152203

src/Page/Form.purs
joneshf-cn: Ah! That makes sense. Thanks for the explanation.

joneshf-cn:
> Not sure if I answered your question?

You did, thanks!

I think it's becoming clearer that going for this untyped version as the thing we operate on all the time wasn't the best decision.

src/Data/Form.purs
davezuch: I'm fine with leaving it. I did figure that was your reason, and I totally agree with it. I was thinking we could make an exception here since this is a case where we're only looking to match one condition, and we know any other condition should fail. But I also think it's valuable to establish the practice of explicitly covering each case when possible, so we can avoid bugs like below.

src/Data/Form.purs
davezuch: I'm not certain we need them to be expressions...we might be better off with
```purescript
data ExprType
  = Boolean Boolean
  | Int Int
  | String String
  | Pair Key ExprType
  | Array (Array ExprType)
```

davezuch: And if it turns out we do need to allow for arrays of expressions, we can add support for that once the use-case becomes clear

src/Data/Form.purs
joneshf-cn: Yeah, that seems fair. Let's wildcard this up!.

src/Data/Form.purs
joneshf-cn: I want to pop up before going down this path. Are we going to implement this for MVP, or are we planning to come back to it afterward? If we're implementing it, we're going to have to also implement `valueRepresentation` (we'll have to traverse the `Expr` to get the value out of it). It doesn't seem _hard_ per say. But it is a non-trivial amount of work.

davezuch: Hmm, why does this force us to implement `valueRepresentation` (side note, need to find a terser name) now? Can't we just render the `name`?

davezuch: I just thought of something. Having `Invalid a` as part of our `InputSource a` coproduct allows us to recover an invalid value if it becomes valid again.

So using my above food example, say `"meal"` is set to `"Breakfast"` and from the current dropdown `"Pancakes"` is selected, so we have `value: UserInput (String "Pancakes")`. The user changes `"meal"` to `"Lunch"` and now we have `value: Invalid (String "Pancakes")` (but the `default` is shown in the dropdown if one exists). Then the user switches back to `"Breakfast"`, suddenly our invalid value is valid again, so we should recover it and end back up with `value: UserInput (String "Pancakes")`.

davezuch:
> I think it's becoming clearer that going for this untyped version as the thing we operate on all the time wasn't the best decision.

Perhaps we should spend some time figuring out a typed system we can use. Maybe after we hit this MVP?

src/Data/Form.purs
joneshf-cn: > Hmm, why does this force us to implement `valueRepresentation` (side note, need to find a terser name) now? Can't we just render the `name`?

I don't think it's a hard requirement,  but whatever work we'd do in PS to grab the `name` out of the `Expr` would be equivalent to us implementing `valueRepresentation` and using it to do the work. Thinking about it a bit more, how are we going to do `ExprType -> Array ExprType` for the dropdown component's [`items`][items]?

If we go back to the impetus for this change, we want options to be `Expr` so we can change the options on the fly. Could we instead have different dropdowns and let the `visibility` decide when it's on the form?

Let's say we have `options :: Array { name :: String, value :: Expr }` or `options :: Array Expr`. and the `Objective` changes, it has to be cased on in the `visibility` for every field,  but they do no further casing on the Objective for `options`, `placeholder`, or `valueRepresentation`. Someone implementing a form has multiple dropdowns for each `Objective`.

For example:

```PureScript
foo :: Field Expr
foo =
  { name: string_ "foo"
  , visibility: equal_ (string_ "objective-foo") (lookup_ "objective" (boolean_ false))
  , description: string_ "foo"
  , key: "foo"
  , input: Dropdown
    { default: Nothing
    , index: 0
    , options:
      [ { name: "foo-1", value: string_ "foo-1" }
      , { name: "foo-2", value: string_ "foo-2" }
      , { name: "foo-3", value: string_ "foo-3" }
      ]
    , placeholder: string_ ""
    , required: boolean_ true
    , value: NotSet
    , valueRepresentation: string_ ""
    }
  }

bar :: Field Expr
bar =
  { visibility: equal_ (string_ "objective-bar") (lookup_ "objective" (boolean_ false))
  , description: string_ "bar"
  , key: "bar"
  , input: Dropdown
    { default: Nothing
    , index: 0
    , options:
      [ { name: "bar-1", value: string_ "bar-1" }
      , { name: "bar-2", value: string_ "bar-2" }
      , { name: "bar-3", value: string_ "bar-3" }
      ]
    , placeholder: string_ ""
    , required: boolean_ true
    , value: NotSet
    , valueRepresentation: string_ ""
    }
  }

baz :: Field Expr
baz =
  { visibility: equal_ (string_ "objective-baz") (lookup_ "objective" (boolean_ false))
  , description: string_ "baz"
  , key: "baz"
  , input: Dropdown
    { default: Nothing
    , index: 0
    , options:
      [ { name: "baz-1", value: string_ "baz-1" }
      , { name: "baz-2", value: string_ "baz-2" }
      , { name: "baz-3", value: string_ "baz-3" }
      ]
    , placeholder: string_ ""
    , required: boolean_ true
    , value: NotSet
    , valueRepresentation: string_ ""
    }
  }
```

Let's say we have `options :: Expr`. If the `Objective` changes, it has to be cased on in the `visibility` for every field (while handling multiple cases), and it has to be cased on within the dropdown (`options`, `placeholder`, and `valueRepresentation`). Someone implementing the form has one large dropdown expression with many cases.

For example:
```PureScript
fooBarBaz :: Field Expr
fooBarBaz =
  { name:
    if_
      (equal_ (string_ "objective-foo") (lookup_ "objective" (boolean_ false)))
      (string_ "foo")
      (if_
        (equal_ (string_ "objective-bar") (lookup_ "objective" (boolean_ false)))
        (string_ "bar")
        (string_ "baz"))
  , visibility:
    if_
      (equal_ (string_ "objective-foo") (lookup_ "objective" (boolean_ false)))
      (boolean_ true)
      (if_
        (equal_ (string_ "objective-bar") (lookup_ "objective" (boolean_ false)))
        (boolean_ true)
        (equal_ (string_ "objective-baz") (lookup_ "objective" (boolean_ false))))
  , description:
    if_
      (equal_ (string_ "objective-foo") (lookup_ "objective" (boolean_ false)))
      (string_ "foo")
      (if_
        (equal_ (string_ "objective-bar") (lookup_ "objective" (boolean_ false)))
        (string_ "bar")
        (string_ "baz"))
  , key: "fooBarBaz"
  , input: Dropdown
    { default: Nothing
    , index: 0
    , options:
      if_
        (equal_ (string_ "objective-foo") (lookup_ "objective" (boolean_ false)))
        (array_
         [ pair_ "foo-1" (string_ "foo-1")
         , pair_ "foo-2" (string_ "foo-2")
         , pair_ "foo-3" (string_ "foo-3")
         ])
        (if_
          (equal_ (string_ "objective-bar") (lookup_ "objective" (boolean_ false)))
          (array_
            [ pair_ "bar-1" (string_ "bar-1")
            , pair_ "bar-2" (string_ "bar-2")
            , pair_ "bar-3" (string_ "bar-3")
            ])
          (array_
            [ pair_ "baz-1" (string_ "baz-1")
            , pair_ "baz-2" (string_ "baz-2")
            , pair_ "baz-3" (string_ "baz-3")
            ]))
    , placeholder: string_ ""
    , required: boolean_ true
    , value: NotSet
    , valueRepresentation: string_ ""
    }
  }
```

The former seems easier for us and for our users. Did I oversimplify things? If that does work, it also seems like it simplifies typeahead a bit as well, since it would be an `Array Expr` as well and would fall under the same rules.

[items]: https://github.com/citizennet/purescript-ocelot/blob/master/src/Components/Dropdown/Component.purs#L30

joneshf-cn: For audit, we discussed live and I did over simplify it. There are some cases we have where it makes sense to have one dropdown with repeated casing. We're going on with the `options :: Expr`.

joneshf-cn:
@davezuch I think I've addressed all the things we talked about. Mind giving it another once-over?

src/Data/Form.purs
```diff
@@ -127,6 +137,121 @@ instance decodeInputSource :: (DecodeJson a) => DecodeJson (InputSource a) where
...
+      Dropdown input ->
+        field { input = Dropdown input { value = UserInput (Val val) } }
+      Text input ->
+        field { input = Text input { value = UserInput (Val val) } }
+      Toggle input ->
+        field { input = Toggle input { value = UserInput (Val val) } }
+    | otherwise = field
+
+setVisibility :: forall a. Key -> a -> Page a -> Page a
+setVisibility key visibility page =
```
davezuch: This doesn't seem to be used anymore, can we delete?

src/Data/Form.purs
```diff
@@ -127,6 +137,121 @@ instance decodeInputSource :: (DecodeJson a) => DecodeJson (InputSource a) where
...
+  setField :: Field Expr -> Field Expr
+  setField field
+    | key == field.key = case field.input of
+      Dropdown input ->
+        field { input = Dropdown input { value = UserInput (Val val) } }
+      Text input ->
+        field { input = Text input { value = UserInput (Val val) } }
+      Toggle input ->
+        field { input = Toggle input { value = UserInput (Val val) } }
+    | otherwise = field
```
davezuch, line 241: Was thinking of how to make this more future-safe. Maybe rewrite as:
```purescript
setField field = if key == field.key
  then case field.input of
     Dropdown input ->
        field { input = Dropdown input { value = UserInput (Val val) } }
      Text input ->
        field { input = Text input { value = UserInput (Val val) } }
      Toggle input ->
        field { input = Toggle input { value = UserInput (Val val) } }
  else field
```
(Not married to the formatting though if you have preferences of when to put things on a new line 😉)

src/Page/Form.purs
```diff
@@ -33,16 +40,26 @@ data Query a
...
 
 type ParentInput = String
 
+type ChildQuery m = Coproduct1 (DropdownQuery m)
+
+type ChildSlot = Either1 DropdownSlot
+
+type DropdownQuery m = Dropdown.Query Query ExprType m
+
+type DropdownSlot = Key
```
davezuch: I don't feel strongly about this, but we could do away with the alias and just use `Key` directly for all our slots instead of aliasing it for every component we add
```purescript
type DateTimeSlot = Key
type DropdownSlot = Key
type SingleTypeaheadSlot = Key
type MultiTypeaheadSlot = Key
```
🤷‍♂️

src/Page/Form.purs
```diff
@@ -191,8 +150,24 @@ component =
...
-  renderInput :: Field ExprType -> H.ComponentHTML Query
+  renderInput :: Field ExprType -> H.ParentHTML Query (ChildQuery m) ChildSlot m
   renderInput (field) = case field.input of
+    Dropdown dropdown ->
+      HH.slot'
+        cp1
+        field.key
+        Dropdown.component
+        { selectedItem: getValue dropdown
+        , items: fold (toArray dropdown.options)
```
davezuch, line 134: Have we tested that when the `options` change, that's reflected in the component? We might need to send a query down to the component telling it to update its items.

davezuch: Which, would seem like a design flaw really...

src/Data/Form.purs
joneshf-cn: I might be missing something, but that seems to be the same as the current code. Sorry for being dense, but which parts change?

src/Page/Form.purs
joneshf-cn: Nope, no tests for that. How do we tend to test Halogen components?

joneshf-cn: Actually, what is it that we lack confidence in? Is it our eval, `ocelot`, or `halogen`? The first seems much easier to validate than the other two. The other two also don't seem like they should need to be validated in `lynx`.

src/Data/Form.purs
davezuch: Ah no, I'm the dense one here. Upon first review I thought the casing was inside the guard, which would mean it didn't have to be total and new `Input` types could fall through to the `otherwise` case. I misread, kindly ignore.

joneshf-cn: No worries.

src/Page/Form.purs
davezuch: So, if I remember correctly, updates to `dropdown.options` in the parent state, don't trigger a new `input` being sent to the child component, we have to manually do it through a query in our parent `eval`. If that's the case it's a Halogen thing, but the design flaw could be considered either in Halogen or Ocelot. Or both.

davezuch: So I manually tested this with PR #9, and they don't update.

davezuch: Your PR is proof that this is a design flaw (and it's bitten us before as well), because the behavior isn't expected. I feel Ocelot should be updated with a `receiver` to handle updates to state

joneshf-cn: Implemented this in https://github.com/citizennet/purescript-lynx/pull/8/commits/c2274913c2ddb22a786034ab112de66a7c195a5b

joneshf-cn:
Okay, I think we're up to date again. Lemme know if there's anything else.

src/Data/Expr.purs
```diff
@@ -5,55 +5,88 @@ import Prelude
...
 import Test.QuickCheck (class Arbitrary, arbitrary)
-import Test.QuickCheck.Arbitrary (genericArbitrary)
-import Test.QuickCheck.Gen (Gen, Size, oneOf, sized)
+import Test.QuickCheck.Gen (Gen, Size, arrayOf, oneOf, sized)
 
 type Key = String
 
 data ExprType
-  = Boolean Boolean
+  = Array (Array ExprType)
```
davezuch, line 19: 🤔 This gives us heterogeneous arrays...until we go back to typed expressions at least

src/Page/Form.purs
```diff
@@ -206,26 +154,39 @@ component =
...
+  EvalForm expr a -> a <$ do
+    { values } <- H.get
+    let evaled' = Lynx.Data.Form.eval (\key -> Data.Map.lookup key values) expr
+    evaled <- for evaled' \page -> do
+      for_ page.contents \section -> do
+        for_ section.contents \field -> do
+          case field.input of
+            Dropdown dropdown ->
+              for_ (toArray dropdown.options) \options ->
+                H.query' cp1 field.key (Dropdown.SetItems options unit)
```
davezuch, line 175: Ooph, sorry to hit you with another one. We're not currently handling the case where an invalid option is selected. So if you load up the current form, select "Strawberry" from the food dropdown, then toggle "Active" to off, "Strawberry" remains selected despite being no longer valid.

I would say the behavior here hasn't really been specified, so I'm okay if you want to merge this in and then tackle this behavior in another PR.

But here the value on the form state should probably become `Invalid (String "Strawberry")`, and the value in the dropdown should be reset to `Nothing`. Glad to discuss more next week.

davezuch (approved):
Looks super good! There's still one gotcha that I mentioned in a separate comment, but I think we can address it in a separate PR and get this one merged in. Will leave that to you!

src/Page/Form.purs
joneshf-cn: Since we don't do any validation for any of the fields, it seems like we should try to tackle that separately. I think we should punt on this.

PR #9, opened 2019-03-08
Add dynamic options to food dropdown

davezuch:
## What it does

Added dynamic options to food dropdown to test if they change in form.


davezuch:
Which, they don't.

davezuch:
You just mentioned you're writing a test for this use case. Feel free to delete this PR if you'd rather cover it in your test

joneshf-cn (approved):
> You just mentioned you're writing a test for this use case. Feel free to delete this PR if you'd rather cover it in your test

No way. Let's add this! This is a good thing to see happening.

PR #10, opened 2019-03-13
PS-2957 Add currency field

joneshf-cn:
## What it does

Adds support for a currency field.

## Review asks

Probably best to go commit by commit, since there's a bit of cleanup before doing the actual work.

### DIFF
--- a/src/Data/Form.purs
+++ b/src/Data/Form.purs
@@ -11,8 +11,9 @@ import Data.Generic.Rep.Show (genericShow)
 import Data.Map (Map)
 import Data.Map as Data.Map
 import Data.Maybe (Maybe(..))
+import Data.Newtype (wrap)
 import Data.Traversable (class Traversable, sequenceDefault, traverse)
-import Lynx.Data.Expr (EvalError, Expr(..), ExprType(..), Key, boolean_, evalExpr, if_, lookup_, string_)
+import Lynx.Data.Expr (EvalError, Expr(..), ExprType(..), Key, boolean_, cents_, evalExpr, if_, lookup_, string_)
 import Test.QuickCheck (class Arbitrary)
 import Test.QuickCheck.Arbitrary (genericArbitrary)
 import Type.Row (type (+))
@@ -56,6 +57,11 @@ type StringRows f r =
   | r
   )
 
+type CurrencyRows f r =
+  ( placeholder :: f
+  | r
+  )
+
 type DropdownRows f r =
   ( options :: f
   , placeholder :: f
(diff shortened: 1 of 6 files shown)


src/Data/Form.purs
```diff
@@ -56,14 +57,20 @@ type StringRows f r =
...
+
 type DropdownRows f r =
   ( options :: f
   , placeholder :: f
   | r
   )
 
 data Input f
-  = Dropdown (Record (SharedRows f + RequiredRows f + DropdownRows f + ()))
+  = Currency (Record (SharedRows f + RequiredRows f + CurrencyRows f + ()))
```
davezuch, line 72: I suppose we might want a `NumberRows` similar to `StringRows` for numeric validation, though that won't be relevant until my PR is done so maybe I should just add those in.

```purescript
type NumberRows f r =
  ( min :: f
  , max :: f
  | r
  )
```

src/Page/Form.purs
```diff
@@ -125,6 +125,15 @@ component =
 
   renderInput :: Field ExprType -> H.ParentHTML Query (ChildQuery m) ChildSlot m
   renderInput (field) = case field.input of
+    Currency currency ->
+      Input.currency_
+        [ HP.id_ field.key
+        , HP.placeholder (print currency.placeholder)
+        , HP.value $ fold do
+          value <- getValue currency
+          _ <- toCents value
```
davezuch, line 134: Hmm not sure I understand this, why is this being discarded?

src/Data/Form.purs
joneshf-cn: I'm down to wait. I didn't want to add fields we weren't using. It's tough to know what all is essential to things have an effect in a real form and what is waiting to be implemented when there are excess fields.

davezuch: They certainly aren't essential to the MVP, and we've already decided not to support all behaviors from the start. So we just need to be mindful when scoping out future sprint work that it's not only fields we need to add support for, but behaviors as well.

At any rate, since this is already done I'm okay with merging this PR in and then adding in validation as part of my PR. Or a new PR that's dependent on both.

src/Page/Form.purs
joneshf-cn: Sorry about that. This line might give us a `Cents` value. But we don't actually want that value, we want the `print`ed representation of the `ExprType`–the line below–if it exists. Does that make sense?

If we took the `Cents` value and rendered it to a `String`, we'd be spreading the logic of how we want to render from `print` to `Lynx.Page.Form`. It seemed better to not open that can of worms.

More than happy to restructure this whole thing so it's clearer if you've any suggestions.

davezuch: I guess I just don't know why we're calling `toCents` at all if we're not using the result. Or because we're in a `Maybe` monad, are we just using this line to validate our value before rendering? That's pretty clever

davezuch (approved):
tenor-1

PR #12, opened 2019-03-15
Speed up CI

joneshf-cn:
## What it does

Our CI times were kind of crazy. We had a median time of 17min 41s. This was causing a real problem as our builds were starting to time out (see [53][] and [55][]) since there was no output.

The main culprit was that `purs` was getting the wrong information about the number of cores it has access to. I think the machines we run on CircleCI have 32 cores, so that's what `purs` was trying to parallelize compilation at. Since we only have access to two cores, we ended up with a ton of thrashing instead of compilation. See: https://github.com/purescript/purescript/issues/1880 for more information.

The fix there was to explicitly call out how many threads to use (`+RTS -N2 -RTS`). The first commit has a better explanation. This fix dropped about seven minutes from build times. So we're about 10 min for a build. That's enough to pass under the timeouts since there will be output during that time.

The secondary culprit was that `pulp` seemed to want to always do a full compile whenever it was asked to do anything. That would be fine if it was also always passing along the RTS options. For whatever reason, it doesn't look like `browserify` or `test` do that. Rather than debug all that stuff and still end up with `pulp` always forcing a full recompile, we drop `pulp` in favor of using `browserify`, `node`, and `purs` directly (really `psa`, but it's a thin wrapper).

Since our `Makefile` is a bit more granular than `pulp`, we for sure  don't have to always dish out to `purs` whenever anything changes. But, we also can give `purs` the RTS options all of the time. This fix dropped an additional eight minutes from build times. So we're about 1 min 30 s for a build (most of it is tests running). That's means CI becomes quick enough that you don't really get to context switch when developing.

## Review asks

Are we okay with ditching `pulp` and using `browserify`, `node`, and `purs` directly to get build times down to 1min 30s on CI (and quicker in development)? I don't feel strongly enough about `pulp` to want to deal with its slowness, but I understand if it's worth it for other people and we want to keep it.

[53]: https://circleci.com/gh/citizennet/purescript-lynx/53
[55]: https://circleci.com/gh/citizennet/purescript-lynx/55


davezuch (approved):
Awesome work, @joneshf-cn! That's a huge speed improvement in build time!

joneshf-cn:
Hmm, the merge didn't run on CI due to the [incident][] they had earlier today. Going to push a commit to force it to run (to make sure things won't break here).

[incident]: https://status.circleci.com/incidents/8547mbrg3p9h

PR #13, opened 2019-03-20
Break apart the expressions

joneshf-cn:
## What it does

Splits the expression type apart in an effort to regain type information in our expressions.  We haven't gotten the whole way there, but we've got other work to do, and the commit we created the PR with is a good stopping point for the changes in here.

The approach we take is the ["Data types à la carte"][a la carte] approach: define all the cases independently and smoosh them together later. We don't go full type class on it (since defining type classes correctly is very hard), so we'll not have fully solved the expression problem with this solution. That is, if we add or alter a case, we might have to go adjust some other case or the combination of them.

The end game of these changes is to write something like:

```PureScript
type StringRows f r
  = ( maxLength :: Inject IntF f => f
    , minLength :: Inject IntF f => f
    , placeholder :: Inject StringF f => f
    | r
    )
```

There's still more work to be done, but that's the hope.

## Review asks

* Sorry for the large PR. There are some pretty sweeping changes here. Might be easiest to go commit by commit. I tried to keep them fairly isolated.
*  Are we okay with trying this approach? I think there are some pretty nice benefits to this: individual cases show where the problems exist–`evalIfF` and `evalEqualF` are the issues with typing–separating the code seems a bit easier to understand in isolation. But, am I too close to the changes? It's possible I'm blind to some negatives.
* It seems like we can get to where we want (static typing of expressions) with some more work. I'm not entirely sure of the path to get there but it seems like if we keep following and pushing on this, we'll make it. Can you see the path clearly?
* We changed the JSON format for `ExprType`. It went from `param :: Json` to `params :: Array Json`. This was done for uniformity–`Expr` and `ExprType` being separate is an implementation detail not an essential part of expressions–Is that okay? I don't think we have any actual consumers at the moment, so it seems fine to break things.

[a la carte]: http://www.cs.ru.nl/~W.Swierstra/Publications/DataTypesALaCarte.pdf


joneshf-cn:
Also, the diff looks huge because of `dist/main.js`. The important part of the diff is:
```
$ git diff --stat origin/master -- ':(exclude)dist/main.js'
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ STDIN
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │  bower.json                        |   3 +-
   2   │  src/Data/Expr.purs                | 339 ++++++++++++++-------------------------------------------
   3   │  src/Data/Form.purs                |  64 ++++++-----
   4   │  src/Lynx/Data/Equal.purs          | 104 ++++++++++++++++++
   5   │  src/Lynx/Data/ExprType.purs       | 425 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
       │ ++++++++
   6   │  src/Lynx/Data/If.purs             |  83 ++++++++++++++
   7   │  src/Lynx/Data/Lookup.purs         |  73 +++++++++++++
   8   │  src/Lynx/Data/Print.purs          |  60 +++++++++++
   9   │  src/Lynx/Data/Val.purs            |  57 ++++++++++
  10   │  src/Page/Form.purs                |  25 ++---
  11   │  test/Test/Data/Expr.purs          |  22 +---
  12   │  test/Test/Lynx/Data/ExprType.purs |  67 ++++++++++++
  13   │  test/Test/Lynx/Data/Form.purs     |  20 ++--
  14   │  test/Test/Main.purs               |   2 +
  15   │  14 files changed, 1014 insertions(+), 330 deletions(-)
───────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────
```

joneshf-cn:
I'm actually going to redo the ideas in here. This PR is too big for what it's intended to do. Will split it into smaller, more relevant PRs.

PR #14, opened 2019-03-21
PS-2968 Add `UserCleared` input

joneshf-cn:
## What it does

As discussed on slack, there is another state to handle: deleting all input from a field. It's a distinct state from `NotSet`, so we handle it accordingly.

Let me know if I missed the mark on this one.


src/Data/Form.purs
```diff
@@ -246,7 +251,7 @@ getValue
   -> Maybe a
 getValue x = userInput x.value <|> x.default
 
-setValue :: Key -> ExprType -> Page Expr -> Page Expr
+setValue :: Key -> InputSource ExprType -> Page Expr -> Page Expr
```
davezuch, line 254: This looks exactly the same as the work I did, except I made this `InputSourc Expr` instead of `ExprType`...looks like yours takes advantage of `InputSource` being a functor to wrap the value with `Val`, if there as a value. I like it, down to keep your way, it's more ergonomic

davezuch (approved):
Down to merge this in, then worry about conflicts from my validation branch

src/Data/Form.purs
joneshf-cn: Yeah, I felt pretty good that we had nearly identical solutions. It affirmed that I hadn't done the wrong thing 😄.

PR #15, opened 2019-03-21
PS-2968 Add DateTimes

joneshf-cn:
## What it does

The change here give us support for DateTimes. We're blocked on finishing this by having `UserCleared` exist. #14 is one way to get unblocked.

## Review asks

Did I miss anything? This seemed to be quite a bit easier to implement than currency or dropdowns.


src/Data/Form.purs
```diff
@@ -184,6 +193,19 @@ eval get page = do
           , value
           }
         )
+    DateTime input -> do
+      default <- traverse (evalExpr get) input.default
+      placeholder <- evalExpr get input.placeholder
+      required <- evalExpr get input.required
+      value <- traverse (evalExpr get) input.value
+      pure
```
davezuch, line 206: An example of the problem with my validation PR, merging that in after this, we'd easily miss that this isn't being validated

src/Page/Form.purs
```diff
@@ -227,3 +247,21 @@ eval = case _ of
     Dropdown.Emit x -> a <$ eval x
     Dropdown.Selected val -> eval (UpdateKey key val a)
     Dropdown.VisibilityChanged _ -> pure a
+
+  DateTimePickerQuery key message a -> case message of
+    DateTimePicker.DateMessage _ -> pure a
+    DateTimePicker.SelectionChanged (Just val) ->
+      eval (UpdateKey key (datetime_ val) a)
+    DateTimePicker.SelectionChanged Nothing -> do
+      let _ = todo $ SProxy :: SProxy "Handle the removal of a datetime. Will require `UserCleared`."
```
davezuch: This is neat, I like it

davezuch (approved):
Looks good! None of my comments are important

PR #11, opened 2019-03-14
PS-2961 Create MVP Form

joneshf-cn:
## What it does

This should meet the layout requirements of V1. It has name, objective, and media budget.

## Review asks

Is this all looking okay? Do we need anything else on the form?

<details>
<summary>V1</summary>

Screen Shot 2019-03-14 at 3 23 27 PM
</details>

PR #16, opened 2019-03-22
Validation

davezuch:
## What it does

This PR adds in support for validating required fields, as well as validating that selected values are valid, when applicable. Introduces a number of new types to accomplish that, and moves some logic around as well. I'll annotate the things I think deserve attention or consideration tomorrow, but wanted to get this out tonight.

## Review asks

Do these changes seem sensible?

## Other Notes

CircleCI build is broken, seems it's not pulling in the latest version of Ocelot despite me busting the cache, I'll work on that more tomorrow as well.


src/Data/Expr.purs
```diff
@@ -107,7 +109,7 @@ print :: ExprType -> String
 print = case _ of
   Array x -> "[" <> intercalate ", " (map print x) <> "]"
   Boolean x -> show x
-  Cents x -> formatCentsToStrDollars x
+  Cents x -> replace (Pattern ".00") (Replacement "") (formatCentsToStrDollars x)
```
joneshf-cn, line 123: Sorry, I'm not really understanding the motivation behind this change. Why are we removing `".00"` if it exists?

src/Data/Form.purs
```diff
@@ -41,6 +43,7 @@ type Field f = Record (FieldRows f ())
 type SharedRows f r =
   ( default :: Maybe f
   , value :: InputSource f
+  , errors :: Errors ValidationError
```
joneshf-cn, line 50: Hmm, adding a field here seems like we're going to have to be diligent when handling errors. Which is to say, it's incredibly likely we're going to add some new field or some new `ExprType` and forget to validate them, or use a new component and not display the errors for it.

Consuming products (records in this case) means that `purs` won't force us to handle all parts of the product and will forget that at some point in the future. The same issue doesn't arise with sums. We would have to actively ignore a case when consuming sums. In other words, we cannot forget to handle errors if they're a sum.

How would you feel about changing `SharedRows _ _` to something like?
```PureScript
type SharedRows f r =
  ( default :: Maybe f
  , value :: Data.These.These (Errors ValidationError) (InputSource f)
  | r
  )
```
We couldn't forget to display an error or handle a new `ExprType` this way.

joneshf-cn: Actually, what's the intent behind having `Errors ValidationError` and also the `Invalid` case of `InputSource`? If we stick with the product approach, it seems like we're introducing some new states that shouldn't exist:
```PureScript
{ errors: singletonError Required
, value: UserInput (boolean_ true)
, …
}
```
```PureScript
{ errors: mempty
, value: Invalid (boolean_ true)
, …
}
```
Neither of those seem like states we should be able to get into.

If we go with the sum approach, we've got a similar situation:
```PureScript
{ value: Data.These.Both (singletonError Required) (UserInput (boolean_ true))
, …
}
```
and
```PureScript
{ value: Data.These.That (Invalid (boolean_ true))
, …
}
```
Seem to represent similar ideas (the input is bunk), but two different ways of expressing it.

It seems like we should change `InputSource _` to represent our new states instead of adding another field to `RequiredRows _ _`.

src/Page/Form.purs
```diff
@@ -206,14 +211,19 @@ eval = case _ of
...
-      _ -> pure a
+      Success { expr } -> do
+        case val of
+          Nothing -> do
+            H.modify_ _ { values = Data.Map.delete key values }
+            eval (EvalForm (Lynx.Data.Form.setValue key UserCleared expr) a)
+          Just v -> do
+            H.modify_ _ { values = Data.Map.insert key v values }
+            eval (EvalForm (Lynx.Data.Form.setValue key (UserInput $ Val v) expr) a)
+      _ -> pure a -- should this be possible?
```
joneshf-cn: Responding to the comment, yes the other three cases of `RemoteData _ _` are all valid here. If we dont' want to handle them in `UpdateKey`, we need to switch our state from:
```PureScript
type State
  = { form :: RemoteData EvalError { expr :: Page Expr, evaled :: Page ExprType }
    , values :: Map Key ExprType
    }
```
to
```PureScript
type State
  = { expr :: Page Expr
    , evaled :: Page ExprType
    , values :: Map Key ExprType
    }
```
I'm not sure how we'd do that an encapsulate the logic of loading data effectfully.

src/Data/Expr.purs
davezuch, line 123: This arguably should be handled in Ocelot, but didn't want to tag another release just for this small feature. This prevents rounded dollar amounts in the `Currency` fields from being formatted with an extra ".00"

src/Page/Form.purs
joneshf-cn: Also, now that we're using `purescript-remotedata@4.1.0` we can remove the explicit casing to make the algorithm a bit clearer:

```PureScript
for_ form \{ expr } -> case val of
  Nothing -> do
    H.modify_ _ { values = Data.Map.delete key values }
    eval (EvalForm (Lynx.Data.Form.setValue key UserCleared expr) a)
  Just v -> do
    H.modify_ _ { values = Data.Map.insert key v values }
    eval (EvalForm (Lynx.Data.Form.setValue key (UserInput $ Val v) expr) a)
```

src/Page/Form.purs
```diff
@@ -185,45 +187,51 @@ component =
...
   Initialize a -> a <$ do
-    H.modify_ _ { values = Lynx.Data.Form.keys testPage }
     void $ bitraverse_
       (\e -> H.modify _ { form = Failure e })
       (\f -> eval $ EvalForm f a)
       (Right testPage)
 
   EvalForm expr a -> a <$ do
-    { values } <- H.get
+    let values = Lynx.Data.Form.keys expr
```
joneshf-cn, line 274: I think if we're going to traverse the `Page Expr` to recompute the values on each `EvalForm` instead of incrementally changing it on `Updatekey`, we don't want to also keep the computed `Map Key ExprType` in the `State`. Which is to say, if carrying around this `Map Key ExprType` and incrementally updating it in `UpdateKey` is causing bugs, we should remove it from the `State`.

src/Data/Form.purs
```diff
@@ -151,6 +161,74 @@ userInput = case _ of
   UserInput x -> Just x
   _ -> Nothing
 
+newtype Errors e = Errors (Set e)
```
davezuch, line 181: I'm not sure `Set e` is the most appropriate type to model a collection of errors. The idea was that since it's a Set, you wouldn't be able to have duplicate errors (e.g. as opposed to an array, where you could have `[Required, Required]`). 

But if my `e` is something like `data ValidationErrors = Required | MaxLength Int`, I can still have `fromFoldable (MaxLength 10 : MaxLength 20 : Nil)`, though that should be pretty hard to get from how our forms work.

Still, it's the best I could come up with for modeling them 🤷‍♂️

src/Data/Form.purs
```diff
@@ -304,27 +307,39 @@ eval get page = do
...
     :: ∀ r
      . Record (SharedRows ExprType + ( required :: ExprType | r ))
     -> Errors ValidationError
-  validateRequired input = do
+  validateRequired input =
     if input.required == Boolean true && isEmpty (getValue input)
       then singletonError Required
       else mempty
 
+  validateInvalid
```
joneshf-cn, line 379: I think this kind of goes back to the point about having too many cases with the `errors` as product approach. We have to stay diligent to do this conversion and not forget it. When we add another field with its own rules for validation, we're going to have to be diligent to remember to convert `Invalid foo` into some other error.

While I recognize the changes here work and it looks like they do what they're supposed to, it seems very easy for us to slip up and introduce bugs.

src/Data/Form.purs
```diff
@@ -176,50 +254,91 @@ eval get page = do
...
-          , value
-          }
-        )
+      pure $ validate $ Currency
+        { default
+        , placeholder
+        , required
+        , value
+        , errors: mempty
+        }
```
davezuch, line 280: So right now I construct a valid `Input ExprType`, then run it through the `validate` function which adds on validation. My issue with this is that the compiler doesn't enforce validation. You could easily forget to add the call to `validate` here and the compiler wouldn't complain. I'm wondering if we should/could add another `f` that represents a validated form, and have `errors` only be a label on `Input`s for that `f` if that's possible. Something like

```purescript
toggleExpr :: Input Expr
toggleExpr = Toggle
  { default: Just (Val $ Boolean true)
  , value: NotSet
  }

toggleExprType :: Input ExprType
toggleExprType = Toggle
  { default: Just (Boolean true)
  , value: NotSet
  }

toggleValidated :: Input Validated
toggleValidated = Toggle
  { default: Just (Boolean true)
  , value: NotSet
  , errors: mempty
  }
```

src/Data/Form.purs
```diff
@@ -176,50 +254,91 @@ eval get page = do
...
-          , placeholder
-          , required
-          , value
-          }
-        )
+      value' <- traverse (evalExpr get) input.value
+      let arrayOptions = Data.Maybe.fromMaybe [] (toArray options)
+          value = case value' of
+            UserInput x ->
+              if (not $ x `Data.Array.elem` arrayOptions)
```
davezuch, line 302: @joneshf-cn I'm not 💯 sold on fully namespacing imported functions like this. I do agree with people on Slack who feel like it adds visual noise, though I also agree it can be useful for clarity. But I also feel `Array.elem` would be just as useful here. It also seems hard to enforce, or consistently do. But since I've seen you use it here and there in this project, I figured I'd test-drive it, see if I don't come around 🙂

.circleci/config.yml
```diff
@@ -20,10 +20,10 @@ jobs:
 
       - restore_cache:
           keys:
-            - v0.1-bower-cache-{{ .Branch }}-{{ checksum "bower.json" }}
-            - v0.1-bower-cache-{{ .Branch }}
-            - v0.1-bower-cache-master-{{ checksum "bower.json" }}
-            - v0.1-bower-cache-master
+            - v0.2-bower-cache-{{ .Branch }}-{{ checksum "bower.json" }}
```
joneshf-cn, line 23: Looking at the build before this commit: https://circleci.com/gh/citizennet/purescript-lynx/81, it looks like the bower cache is working correctly. The issue seems to be the `output` caching. My guess is that the timestamp of `output/Ocelot.Block.FormField/index.js` is newer than the timestamp of `bower_components/purescript-ocelot/src/Blocks/FormField.purs` even though the contents of `bower_components/purescript-ocelot/src/Blocks/FormField.purs` are different from the file contents that built `output/Ocelot.Block.FormField/index.js`. It's this issue rearing its head again: https://github.com/purescript/purescript/issues/3145.

The cache keys were changed in https://github.com/citizennet/purescript-lynx/pull/11/commits/255a943e45161b1818275a5ca05631feff2cb5b5. I don't think it needs to stay this way, so feel free to change to a different set of keys if it'll mask the issue.

src/Page/Form.purs
```diff
@@ -100,21 +102,24 @@ component =
...
+            [ Card.card_ [ Format.p [ css "text-red" ] [ HH.text "Loading..." ] ] ]
+          Failure e ->
+            [ Card.card_ [ Format.p [ css "text-red" ] [ renderEvalError e ] ] ]
+          Success page ->
+            append
+            [ Format.heading_
+              [ HH.text page.evaled.name ]
+            ]
+            $ renderSection <$> page.evaled.contents
+      , HH.pre_ [ HH.text $ show values ]
```
davezuch, line 135: Added a little debug feature here, rendering out the current form values so that we can be sure our form's in the correct state. Would be nice if we could pretty print this

src/Page/Form.purs
```diff
@@ -158,12 +163,10 @@ component =
           value <- getValue currency
           _ <- toCents value
           pure (print value)
+        , HE.onValueChange (HE.input $ UpdateValue field.key <<< Data.Maybe.maybe UserCleared (UserInput <<< Val <<< Cents) <<< parseCentsFromDollarStr)
```
davezuch: This line got kind of gnarly 😅

davezuch: I moved the `InputSource` from `setValue` to here so we can choose a value other than `UserInput` if it makes sense to. Looks like you had the same idea in you `UserCleared` PR @joneshf-cn

src/Page/Form.purs
```diff
@@ -177,53 +180,80 @@ component =
...
-        [ HP.value $ fromMaybe "" $ toString =<< getValue input
+        [ HP.value $ Data.Maybe.fromMaybe "" $ toString =<< getValue input
         , HP.placeholder $ print input.placeholder
         , HP.id_ field.key
+        , HE.onValueChange (HE.input $ UpdateValue field.key <<< UserInput <<< Val <<< String)
         ]
     Toggle input ->
       Toggle.toggle
-        [ HP.checked $ fromMaybe false $ toBoolean =<< getValue input
+        [ HP.checked $ Data.Maybe.fromMaybe false $ toBoolean =<< getValue input
```
davezuch, line 213: Feels weird to have a default here for when no default is specified. I feel like that means we would have `getValue toggle == Nothing` but on the form it would appear to be `false`. Maybe we need support in Ocelot for a toggle that isn't set. A `Maybe Boolean` toggle perhaps 🤔

src/Page/Form.purs
```diff
@@ -177,31 +180,56 @@ component =
...
-      H.modify_ _ { values = Lynx.Data.Form.keys page }
-      eval (EvalForm page a)
+    bifor_ page'
+      do \e -> H.modify _ { form = Failure e }
+      do \f -> eval (EvalForm f a)
 
   EvalForm expr a -> a <$ do
-    { values } <- H.get
-    let evaled' = Lynx.Data.Form.eval (\key -> Data.Map.lookup key values) expr
+    let values = Lynx.Data.Form.keys expr
```
davezuch, line 274: Moved this here, instead of doing it both on `Initialize`, and then updating it from `UpdateValue`. Seemed like an easier way to keep `values` consistent

src/Page/Form.purs
```diff
@@ -76,21 +78,24 @@ component =
...
+            [ Card.card_ [ Format.p [ css "text-red" ] [ HH.text "Loading..." ] ] ]
+          Failure e ->
+            [ Card.card_ [ Format.p [ css "text-red" ] [ renderEvalError e ] ] ]
+          Success page ->
+            append
+            [ Format.heading_
+              [ HH.text page.evaled.name ]
+            ]
+            $ renderSection <$> page.evaled.contents
+      , HH.pre_ [ HH.text $ show values ]
```
joneshf-cn, line 135: Making a comment here, so if it disappears feel free to ignore it (I'm going commit by commit). I assume this is for debugging, let's not forget to remove this.

joneshf-cn (commented):
Wooo! Validation!!!!

I've got one qualm with the way we're representing errors. The rest of my comments are less important to me. Let's discuss that and merge afterwards!

joneshf-cn:
Sorry, I'd started a review and not checked back to see your comments. Lemme read through it now.

src/Data/Expr.purs
joneshf-cn: Gotcha! More than happy to merge this now and move to `ocelot` later.

joneshf-cn: Answered in https://github.com/citizennet/purescript-lynx/pull/16#discussion_r268301313

src/Data/Form.purs
joneshf-cn: Seems right to me!

src/Data/Form.purs
joneshf-cn: Yeah, I feel similarly about the lack of static checking for errors. I made some comments about this: https://github.com/citizennet/purescript-lynx/pull/16#discussion_r268286652.

Ultimately, I think we need to decide what `Invalid` means for `InputSource _`. If it's supposed to represent an invalid value, it seems like that's where we should be putting errors. What do you think?

src/Data/Form.purs
joneshf-cn: Heh, no worries. I've tried not to do it too much, because that's not been the style here and I don't feel like I should be changing the style around a bunch. Don't feel pressured to do it just because I'm working on this package. I'm more than happy to follow whatever style we've got.

src/Page/Form.purs
joneshf-cn: Ah, I see. So we want to keep this? I'm game. I wonder if there's something in halogen or a halogen package to do this for us?

src/Page/Form.purs
joneshf-cn: I think this is an artifact of both halogen stuff taking `Array _` as arguments and there being a sensible default for `checked`. It's easier to handle the `Maybe Boolean` inside of the array than it is to handle it on the outside. What we should probably do is decide whether to add the `checked` property based on what `getValue >=> toBoolean` returns. We _should_ be doing this in all places where we use `getValue`.

In other words, I think `ocelot` is fine to keep doing what it's doing. We're not using it the best way.

src/Page/Form.purs
joneshf-cn: I commented on this here: https://github.com/citizennet/purescript-lynx/pull/16#discussion_r268304532. But I'll pull it here for coherency.

I don't have a preference for how we do things, but `UpdateValue` is now pretty anemic and we don't ever use the `values :: Map Key ExprType` in the `State`. I think if we're going to be recomputing the values every `EvalForm`, let's drop `values` from the `State` and kill `UpdateValue`.

src/Page/Form.purs
joneshf-cn: Conversation is in: https://github.com/citizennet/purescript-lynx/pull/16#discussion_r268318862

src/Page/Form.purs
joneshf-cn: Conversation is in: https://github.com/citizennet/purescript-lynx/pull/16#discussion_r268311886.

src/Page/Form.purs
davezuch: For now, I think it's useful. Eventually we'll want to hide it, maybe expose it via an env flag or a `?debug` param in the url like we do for our current forms

src/Page/Form.purs
davezuch: Hmm, that's true, but there's still the issue that even if we don't pass in `checked` at all, the toggle block might not have a render state that represents that. I think it would just treat it as unchecked or `false`

joneshf-cn: Is that not expected?

src/Data/Form.purs
davezuch: @joneshf-cn thanks for pointing these issues out, you're 💯 right. So maybe we merge `ValidationError` in with `InputSource`? The question then is how?
```purescript
data InputSource e a
  = NotSet
  | UserCleared
  | Invalid e a
  | UserSet a
```
Still doesn't seem quite right, as `NotSet` and `UserCleared` could be invalid, depending on whether the field is `required`.

Plus there's the `default` value to think about. I don't know that we can roll validation into `InputSource` without rolling `default` value in as well. Plus `InputSource` is not actually meant to act as validation, but as a means of knowing how the value was set. It was either not set, cleared by the user, cleared by form logic (my original intent was that invalid dropdown values would be cleared, but how I implemented it, they aren't, we just show an error), or set by the user.

Adding in validation changes the meaning and potentially usefulness of `InputSource`. Maybe the solution is separating `InputSource` from any value.

```purescript
data InputSource
  = Form
  | User

data ValidationResult a
  = Valid a
  | Required
  | Invalid a
  | MinLength Int a
  | MaxLength Int a
```
This covers the `NotSet` and `UserCleared` cases, because if they aren't required our `a` would be a monoid (or at least a `Maybe`).

Then we would have an unvalidated version of our `Input` that only has a value, and a validated one.

```purescript
toggleExprType :: Input ExprType
toggleExprType = Toggle
  { default: Just (Boolean true) :: Maybe ExprType
  , required: Just (Boolean true) :: Maybe ExprType
  , source: Form :: InputSource
  , value: Nothing :: Maybe ExprType
  }

toggleValidated :: Input Validated
toggleValidated = Toggle
  { default: Just (Boolean true) :: Maybe ExprType
  , required: Just (Boolean true) :: Maybe ExprType
  , source: Form :: InputSource
  , value: Valid (Boolean true) :: ValidationResult (Maybe ExprType)
  }
```
Though at that point, if we can, maybe we should just get rid of the `default` and `source` labels from our `Input Validated` record, as it's not useful anymore and `default` can yield invalid state.

There are a couple issues. This doesn't allow for multiple validation errors. Though I don't know yet that we need them. `[Required, MinLength 1]` is redundant.

Also, if we adopted this route we'd prob want to make `ValidationResult` a `Variant` instead of sum type, so that we can't have `MinLength 2` on a `Toggle`.

src/Page/Form.purs
davezuch: Maybe? It's kind of weird to have `required` at all on there then. Because if it's set to `true` and no `default` is defined, and the user doesn't ever toggle it to on, they're going to get a very unexpected `Required` error

src/Page/Form.purs
davezuch: Well, it is possible with how our state is modeled. What I really meant is, should it be possible? Since if we don't have a form, how would we have any key/value to update here? But you also mentioned in another comment how after more recent changes, this query really serves no purpose, so I'll likely get rid of it and avoid the problem altogether

src/Data/Form.purs
joneshf-cn: > `NotSet` and `UserCleared` could be invalid, depending on whether the field is `required`.

This line I think says everything we need to know. `required` isn't a `Boolean` that we product onto `InputSource _`. `required` determines what `InputSource _` is. We don't have:

```PureScript
data Input a
  = Currency
      { errors :: Errors ValidationError
      , required :: Boolean
      , value :: InputSource a
      , …
      }
  | Dropdown
      { errors :: Errors ValidationError
      , required :: Boolean
      , value :: InputSource a
      , …
      }

data InputSource a
  = UserInput a
  | Invalid a
  | UserCleared
  | NotSet
```

We have something more like:
```PureScript
data Input a
  = OptionalCurrency
      { value :: OptionalValue a
      , …
      }
  | OptionalDropdown
      { value :: OptionalValue a
      , …
      }
  | RequiredCurrency
      { value :: RequiredValue a
      , …
      }
  | RequiredDropdown
      { value :: RequiredValue a
      , …
      }

data OptionalValue a
  = OptionalMin Int a
  | OptionalMax Int a
  | OptionalInput a

data RequiredValue a
  = NotSet
  | UserCleared
  | RequiredMin Int a
  | RequiredMax Int a
  | RequiredInput a
```
Or whatever.

We can manipulate the data types to be more ergonomic, but what's important is that the former encoding with `required :: Boolean` had `8 + 2a` states for just `Currency`s value, while the suggested one has `2 + 6a` states for just `Currency`s value.

src/Page/Form.purs
joneshf-cn: Removing the `UpdateKey`/`UpdateValue` case won't remove your concern, it'll just push it into `EvalForm`. I think if we don't want to deal with `RemoteData _ _`, we'd have to take a fully realized `Page Expr` and have the parent component deal with `RemoteData _ _`. Both ways seem feasible to me (keeping it in this component vs. pushing to the parent), so I'm not sure which way we should push.

src/Data/Form.purs
joneshf-cn: That's not to say we have to decode/encode directly to that, we can keep the serialization simpler for our consumers, but we should work with a representation that more correctly matches our mental model of what the form is.

Or we can go directly from that to JSON, it's orthogonal to the concern is what I'm trying to say.

davezuch: I felt like this needed a bit of rethinking, so I went over to Notion for that: https://www.notion.so/citizennet/Rethinking-Input-Types-6b87407f5872419ca50730d523745eee

Let me know your thoughts @joneshf-cn

joneshf-cn: I'm sure you've seen, but I left some comments in the notion. Ultimately, I think we should decide on something and move on. Having any sort of validation is better than none, and I don't want this PR to stymie more while trying to find the perfect representation. If you want to explore what you've written, I say let's go for it!

davezuch: @joneshf-cn how do you feel about merging this in and improving it in a separate PR? It's already getting so big that any more changes are going to be difficult to parse out. And it _does_ work. I've merged in `master` and resolved conflicts, so it's good to go.

I do want to address the issues brought up, but the ideas I've drawn up in Notion aren't thought out enough. E.g., turns out we do need something like `NotSet`, so that we can know whether to display validation errors or not. It's possible that can be accomplished another way or with another value, but I think these types still need more thought.

davezuch: I'd also be okay with not merging it in until these issues are addressed. Will just have to periodically keep it up to date with `master`, but that would at least force us to try to answer these issues sooner than later

joneshf-cn: Merge it up!

PR #17, opened 2019-03-26
PS-2970 Implement a naïve typeahead

joneshf-cn:
## What it does

This is only part of PS-2970. It doesn't finish the issue.

This gives us just the bare minimum to have a typeahead. It follows the dropdown approach and just uses arrays of pairs.

## Review asks

* This is intentionally not doing a lot so we can keep the size of our PRs a bit small. It's not doing anything async, it's not parsing JSON, it's not working with arbitrary data. I'd like to push those concerns to other PRs so as to lower the burden when doing reviews in an effort to hopefully make reviewing a bit easier and quicker. Are we okay with that?
* Did we handle the typeahead messages properly? I remember there was an issue with the dropdown messages where we had to do some extra threading. It's not clear to me if we have to do that here as well.


src/Data/Form.purs
```diff
@@ -247,6 +255,11 @@ eval get page = do
       default <- traverse (evalExpr get) input.default
       value <- traverse (evalExpr get) input.value
       pure (Toggle { default, value })
+    Typeahead input -> do
+      default <- traverse (evalExpr get) input.default
```
davezuch, line 258: Having a `default` for typeaheads poses some interesting questions. Can we ensure it has the same shape as API data? Do we need to?

src/Page/Form.purs
```diff
@@ -204,6 +211,27 @@ component =
         , HP.id_ field.key
         , HE.onChecked (HE.input $ UpdateKey field.key <<< UserInput <<< Lynx.Data.Expr.Boolean)
         ]
+    Typeahead typeahead ->
```
davezuch: Hmm, we might want to account for the fact that eventually we'll need to support two different typeaheads: single and multi (`Maybe item` vs `Array item`)

Perhaps we should name this `TypeaheadSingle`

src/Page/Form.purs
```diff
@@ -204,6 +211,27 @@ component =
...
+      HH.slot'
+        cp3
+        field.key
+        Typeahead.single
+        ( Typeahead.syncSingle
+          { itemToObject: \item -> fold do
+            { name: name', value: value' } <- toPair item
+            name <- toString name'
+            value <- toString value'
+            pure (Foreign.Object.singleton name value)
```
davezuch: `singleton` creates a key/value pair, and `Fuzzy` only searches the values, so this means `name` won't be used to search against. This should probably be:
```purescript
Foreign.Object.fromFoldable [ Tuple "name" name, Tuple "value" value ]
```

davezuch:
> Did we handle the typeahead messages properly? I remember there was an issue with the dropdown messages where we had to do some extra threading. It's not clear to me if we have to do that here as well.

Once we turn this into an async typeahead, we'll provide the typeahead with the config it needs to manage its own items, so no extra threading will be necessary.

On the other hand, if we wanted to manage the async ourselves, we could capture the `Typeahead.Searched _ -> pure a` message, manually run our search, then thread the results back down to our sync typeahead.

src/Data/Form.purs
joneshf-cn: No idea. I was going off the Notion and adding as few things as possible. More than happy to remove it if it makes things simpler.

davezuch: I was just thinking out loud. I'm okay to keep it until it actually proves problematic

src/Page/Form.purs
joneshf-cn: I see. That doesn't seem to give us what we want then, right? I assume we're concerned that the `name` and `value` might be different. Isn't it possible that a search would find one and not the other? Also, which one should we want? Should we have an array of strings instead of an array of pairs?

joneshf-cn:
> Once we turn this into an async typeahead, we'll provide the typeahead with the config it needs to manage its own items, so no extra threading will be necessary.

Makes sense!

> On the other hand, if we wanted to manage the async ourselves, we could capture the `Typeahead.Searched _ -> pure a` message, manually run our search, then thread the results back down to our sync typeahead.

It seems like we can let the typeahead do its thing, and we just post-process the results.

src/Page/Form.purs
davezuch: What we're likely to end up with is an array of json, and the user will define the key i.e. `name` or `value` to search through. They'll also define the same for rendering

davezuch: The [docementation for `Data.Fuzzy.match`](https://pursuit.purescript.org/packages/purescript-fuzzy/0.2.1/docs/Data.Fuzzy#v:match) might make it more clear what this `item -> Object String` is for

src/Data/Form.purs
joneshf-cn: Ah. Gotcha!

src/Page/Form.purs
joneshf-cn: Oh, that makes a ton of sense! Thanks for explaining it. I should really read the documentation more 😅.

joneshf-cn: Wait, I remember now. I was going off the assumption that we didn't actually care about the `name`, only the `value`. That's similar to dropdowns, where we don't actually care about the `name`, only the `value`. When we move on to the rest of the features for typeaheads (parsing JSON), we'll only have one field to search through anyway. Won't this will be duplicated information?

joneshf-cn: In any case, I've made the changes. We can deal with the other stuff later.

davezuch: Well we'll want to let the users decide what they care about. For a type such as:
```purescript 
{ firstName :: String 
, last Name :: String
}
```
Most likely we would want to use both  values for search and display

davezuch: Well we'll want to let the users decide what they care about. For a type such as:
```purescript 
{ firstName :: String 
, last Name :: String
```
Most likely we would want to use both  values for search and display

src/Page/Form.purs
```diff
@@ -204,6 +211,27 @@ component =
...
+        cp3
+        field.key
+        Typeahead.single
+        ( Typeahead.syncSingle
+          { itemToObject: \item -> fold do
+            { name: name', value: value' } <- toPair item
+            name <- toString name'
+            value <- toString value'
+            pure (Foreign.Object.fromHomogeneous { name, value })
+          , renderFuzzy: \(Fuzzy { original }) -> HH.text $ fold do
```
davezuch: We'll want to include match highlighting in our results. See Ocelot example:
https://github.com/citizennet/purescript-ocelot/blob/b71ad0e41745a9146119914dac8a21900b511a6e/ui-guide/Components/Typeaheads.purs#L159

davezuch (approved):
Looks solid!

PR #19, opened 2019-03-27
PS-2970 Parse JSON for the typeahead options

joneshf-cn:
## What it does

The changes here add support for parsing arbitrary JSON blobs. The JSON blobs are still hardcoded data, but we can at least get a feel for what it means to parse JSON.

Similar to #17, we're taking a naïve approach here and looking down a path for results and fields. That seems like we'll be able to hit MVP if we do that (the JSON data is a simplified version of what the form takes now).

## Review asks

Nothing in particular, just normal review stuff.


davezuch (approved):

davezuch:
Looks good! I'm liking breaking PR's up into more digestible bits

PR #20, opened 2019-03-29
PS-2967 Load data asynchronously for typeaheads

joneshf-cn:
## What it does

This gives us async behavior for typeaheads! We're loading the same MVP JSON blobs and using that for it at the moment.

## Review asks

* Given this is the first time I've used the async typeahead, I might have missed something. I noticed sometimes we [override the `insertable` field][insertable], but I couldn't figure out what it was for. Do we need to do that here?
* We're pulling the JSON blob from GitHub, will that work for MVP or should we find some other location to load the JSON?

[insertable]: https://github.com/citizennet/purescript-ocelot/blob/3f4de3f30ae57efb483ec3d0c8dfffcfb46bb543/ui-guide/Components/Typeaheads.purs#L187


src/Data/Form.purs
```diff
@@ -460,6 +469,49 @@ setValue key val page = page { contents = map setSection page.contents }
         field { input = TypeaheadSingle input { value = map val_ val } }
     | otherwise = field
 
+parseTypeaheadJSON ::
+  forall r.
+  { results :: ExprType, resultValue :: ExprType | r } ->
+  Json ->
+  Either String (Array ExprType)
```
davezuch: I'd prefer not to have a hodgepodge of different styles for how we do things, which is why I'm working on the style guide. Function signatures is one of those things, the format we've settled on at CN is:
> If a function signature gets too long to fit on one line, drop the `::` to a new line, then the `.` and every top-level arrow should also start a new line, with all arguments aligning with the `∀`.
>```purescript
>reallyLongSig
>   :: ∀ f a b
>    . Functor f
>   => Show a
>   => Show b
>   => (a -> b)
>   => f a
>   -> f (Tuple String String)
If you're not a fan we can definitely open a conversation about changing the convention 🙂

src/Data/Form.purs
```diff
@@ -614,7 +659,8 @@ mvpTargetableInterest =
       { default: Nothing
       , options: val_ (array_ [])
       , resultValue: val_ (array_ [string_ "name"])
-      , results: val_ (array_ [])
+      , results: val_ (array_ [string_ "targetableInterest"])
```
davezuch, line 662: So, does using an array here allow us to look for results in multiple places? That's pretty dope, if I'm understanding correctly

davezuch (approved):
Really like what you came up with for pulling in data with a dynamic source and shape 💯 

Did have a comment about formatting, which, hopefully I'll have the guide done soon and we can refer to that. But yeah, this looks awesome!

src/Data/Form.purs
joneshf-cn: Sorry, I tend to write this way out of habit and try to remember to convert to arrow first before committing. I must've missed a couple. Will convert them.

I do have thoughts on arrow first vs. arrow last, but two points that overrule them:
* consistency is more important
* computers are really good at validating style, so humans shouldn't have to it

src/Data/Form.purs
joneshf-cn: Ah, no sorry, the array of strings is a path to the results, similar for the value. Can we change something to make that clearer?

src/Data/Form.purs
davezuch: I'm open to hearing them, unless you'd rather not.
On your second point, is that a suggestion for auto-formatting?

src/Data/Form.purs
davezuch: Ahh okay, that makes sense. This assumes an object path, then? Which I think is a pretty sane assumption, I don't think we need to concern ourselves with supporting a path like `data[0].results`

davezuch: I'm not sure it needs to be clearer...but I suppose we could have a `path_` which is just an alias for `array_` or something 😅 

Or maybe we just call this label `resultsPath` instead of just `results`

joneshf-cn: It does–because we assume all elements are `String`s–but it seems like we ought to be able to handle that case if we assume elements could be `Int`s or `String`s:
```PureScript
array_ [string_ "data", int_ 0, string_ "results"]
```

davezuch: Ah true, yeah so at least this let's us add in support for that down the road without really changing syntax ✨

src/Data/Form.purs
joneshf-cn: I've got it typed up in an old slack post:

> The grammar of type signatures is elegant in a theoretical sense. But, kind of hard in a practical sense. Lining up the operators on the left looks very nice. I mean, what's not to like about:
> ```
> liftStateAt ∷
>   ∀ q sym s a r
>   . IsSymbol sym
>   ⇒ Row.Cons sym (STATE s) q r
>   ⇒ SProxy sym
>   → State s a
>   → Run r a
> ```
> 
> There's a bit of overhead with doing things this way. In order to understand what the thing is to the right of `.` or `⇒`, you have to read the line below it. For instance, What is `IsSymbol sym`? Well, if you read to the next line, it's a constraint. What is `SProxy sym`? Well, if you read to the next line, it's a type.
> 
> Appending the operators on the right is less common, but makes things a bit simpler:
> 
> ```
> liftStateAt ∷
>   ∀ q sym s a r.
>   IsSymbol sym ⇒
>   Row.Cons sym (STATE s) q r ⇒
>   SProxy sym →
>   State s a →
>   Run r a
> ```
> 
> Now, in order to understand what a thing is, you have to read to past the end of the thing. This is how reading type signatures works on one line. If you want to know what `IsSymbol sym` is, you continue reading forward until you reach an operator (or the end of the signature). If you want to know what `SProxy sym` is, you continue reading forward.
> 
> Reading type signatures with operators on the left requires you to parse two lines and backtrack to the original line in order to understand the original line.
> 
> Reading type signatures with operators on the right requires you to parse one line the same way you'd parse if the entire signature were one line.

There's also the diffing noise that can come from arrow first vs arrow last when altering a type signature. In arrow first, it's much more likely that you'll see two-line diffs (one with the change, one with syntactic noise). In arrow last, you only ever see a one line diff for each argument/constraint/parameter change; if a line changed, it's signal, not noise.

> On your second point, is that a suggestion for auto-formatting?

No necessarily, a linter like `rubocop` or `yamllint` can validate style alone so humans don't have to. Though, many `rubocop` cops do auto-correct violations.

joneshf-cn:
I think I've addressed the changes so I'm going to merge this. More than happy to continue the conversation though!

src/Data/Form.purs
davezuch: Interesting, that's a valid point. I'm gonna propose continuing this conversation on the draft of the style guide once I have a first draft. My goal for that is to open it up to discussion/questioning/collaboration, so that seems the appropriate platform.

PR #21, opened 2019-04-02
Split apart `ExprType`

joneshf-cn:
## What it does

This is some of the work that was in #13, but not so huge of a change all at once. We split the `ExprType` apart in an effort to regain type information in our expressions. We haven't gotten the whole way there, but this is a pretty nice stopping point.

From outside of `Lynx.Data.Expr`, basically nothing has changed. If you're using `array_`/`toArray` and friends, it all feels the same. If you try to case on the internal representation, you're in for some work. Inside of `Lynx.Data.Expr`, most of the changes here are boiler plate to make the types line up (many instances had to be made). Hopefully you can see through the boilerplate of these initial changes.

The approach we take is the ["Data types à la carte"][a la carte] approach: define all the cases independently and smoosh them together later. We don't go full type class on it (since defining type classes correctly is very hard), so we'll not have fully solved the expression problem with this solution. That is, if we add or alter a case, we might have to go adjust some other case or the combination of them.

## Review asks

* Are we okay with trying this approach? I think there are some pretty nice benefits to this and were explained in #13. But, am I too close to the changes? It's possible I'm blind to some negatives.
* It seems like we can get to where we want (static typing of expressions) with some more work. I see the path more clearly now. I think we can have things like:
    ```PureScript
    type Dropdown
      = { options ::
          Free (IfF <\/> EqualF <\/> LookupF) (ArrayF (PairF (StringF _)))
        , ...
        }
    ```

    Or something similar. Where we're saying that we can have any nesting of ifs, equals, and lookups so long as it ends with an array of pairs of strings. Does that make sense?

[a la carte]: http://www.cs.ru.nl/~W.Swierstra/Publications/DataTypesALaCarte.pdf


src/Data/Expr.purs
```diff
@@ -2,89 +2,389 @@ module Lynx.Data.Expr where
...
+
+derive instance functorBooleanF :: Functor BooleanF
+
+derive newtype instance showBooleanF :: Show (BooleanF a)
+
+instance foldableBooleanF :: Foldable BooleanF where
+  foldl f z x = foldlDefault f z x
+  foldr f z x = foldrDefault f z x
+  foldMap f = case _ of
+    Boolean _ -> mempty
```
davezuch, line 85: Can't all of these just be
```purescript
foldMap f _ = mempty
```
Or maybe there's something I'm missing?

src/Data/Expr.purs
```diff
@@ -2,89 +2,389 @@ module Lynx.Data.Expr where
...
+newtype ArrayF a
+  = Array (Array a)
+
+derive instance eqArrayF :: (Eq a) => Eq (ArrayF a)
+
+derive instance eq1ArrayF :: Eq1 ArrayF
+
+derive instance functorArrayF :: Functor ArrayF
+
+derive newtype instance showArrayF :: (Show a) => Show (ArrayF a)
```
davezuch: Can we use `genericShow` here? The newtype instance leaves out the constructor

src/Data/Expr.purs
```diff
@@ -2,89 +2,389 @@ module Lynx.Data.Expr where
...
+derive newtype instance traversableArrayF :: Traversable ArrayF
+
+decodeArrayF :: ExprJSON -> Either String (ArrayF ExprJSON)
+decodeArrayF { out, param } = case out of
+  "Array" -> map Array (decodeJson param)
+  _ -> Left "Expected Array"
+
+encodeArrayF :: ArrayF ExprJSON -> ExprJSON
+encodeArrayF = case _ of
+  Array x -> base "Array" x
```
davezuch, line 59: I already commented on `foldMap` below, but noticing this here too, I'm wondering if there's some motivation for using `case` when there's just one pattern to match on? I feel like these seem more concise

```purescript
encodeArrayF (Array x) = base "Array" x
```
```purescript
encodeArrayF = base "Array" <<< unwrap
```

davezuch: Is it just consistency?

src/Data/Expr.purs
```diff
@@ -2,89 +2,389 @@ module Lynx.Data.Expr where
...
+
+isEmptyArrayF :: forall a. ArrayF a -> Boolean
+isEmptyArrayF = case _ of
+  Array [] -> true
+  Array _ -> false
+
+printArrayF :: ArrayF String -> String
+printArrayF = case _ of
+  Array x -> show x
```
davezuch, line 68: So the point of `print`, at least as I intended, was to cast our types to strings for the purpose of concating, ultimately for the purpose of printing to the screen. My original motivation, when still working with primitives, was something like `show`, but that wouldn't wrap quotes around strings. That way concating `"foo"` and `"bar"` wouldn't produce `""foo""bar""`, but just `"foobar"`.

Extending that idea to arrays, we probably just want a comma-separated string of array values. Something like
```purescript
Array x -> intercalate ", " (print <$> x)
```

src/Data/Expr.purs
```diff
@@ -2,89 +2,389 @@ module Lynx.Data.Expr where
...
+instance foldableBooleanF :: Foldable BooleanF where
+  foldl f z x = foldlDefault f z x
+  foldr f z x = foldrDefault f z x
+  foldMap f = case _ of
+    Boolean _ -> mempty
+
+instance traversableBooleanF :: Traversable BooleanF where
+  sequence = traverse identity
+  traverse f = case _ of
+    Boolean x -> pure (Boolean x)
```
davezuch, line 93: So, one thing I was wondering, I was reading the Data Types paper, and saw when they implemented `Functor` for `Val` and `Add`, they only apply the mapping function for the `Add` instance, but just ignore it on the `Val` instance like you're doing here
```haskell
instance Functor Val where
  fmap f (Val x) = Val x
instance Functor Add where
  fmap f (Add e1 e2) = Add (f e1) (f e2)
```
I'm sure I just haven't gotten far enough, but was wondering why that is. Is that simple to explain or do I just need to keep reading to build up an intuition for it?

src/Data/Expr.purs
```diff
@@ -2,89 +2,389 @@ module Lynx.Data.Expr where
...
+encodePairF = case _ of
+  Pair x -> base "Pair" x
+
+isEmptyPairF :: forall a. PairF a -> Boolean
+isEmptyPairF = case _ of
+  Pair x -> false
+
+printPairF :: PairF String -> String
+printPairF = case _ of
+  Pair x -> show x
```
davezuch, line 284: Should we just print `name` here?

I'm starting to wonder if we even need `Pair` at all. The motivation for it is to be able to have a key/value pair where the value is something human readable, like "Condé Nast", and the value is machine-readable, i.e. a unique ID like `13`.

I'm not sure that's necessary here, as all these values are just input by the user, they don't correspond to any primary key in any database or anything. That sort of thing will live at the mapping layer, where if we replace `Pair` with `String`, the user can just say company `"Condé Nast"` corresponds to companyId `13` on a campaign.

The main value I see in keeping it, is that we can change the name without changing the value and having to propagate our changes. E.g. changing
```purescript
Pair { name: "Conde Nast", value: "conde-nast" }
```
to
```purescript
Pair { name: "Condé Nast Entertainment", value: "conde-nast" }
```
wouldn't require us to change our mapping function.

We could also have things like
```purescript
options: Array
  [ Pair { name: "Announce", value: 1 }
  , Pair { name: "Launch", value: 1 }
  , Pair { name: "Onsale", value: 2 }
  , Pair { name: "Final", value: 2 }
  ]
```
Where multiple options can have the same value and so map to the same result in our campaign. Though I don't expect that to be a common scenario.

davezuch (approved):
This is dope. I haven't fully wrapped my head around the recursions scheme stuff, but let's not wait for me on that one 😅 

Glad we're making progress on moving back to types! This approach seems promising! 🎉

src/Data/Expr.purs
davezuch: Wait, I think I get it

davezuch: In the paper, their example for concrete values is
```purescript 
data Val e = Val Int
type IntExpr = Expr Val 
```
`Expr` being defined as
```purescript 
data Expr f = In (f (Expr f))
```
I can see since the `e` isn't inhabited, there is no recursive subtree, which is what the `f` mapping function operates on  
```purescript 
In (Val x)
```

src/Data/Expr.purs
joneshf-cn: You're not missing anything. I'm a judicious `case`er. If we're going for concision, we could do `foldMap = mempty`.

src/Data/Expr.purs
joneshf-cn: It's more about dealing with the codebase as it changes over time.

With these Haskell-based languages, top-level pattern matches get pretty noisy. It's harder to tell at a glance what is important and what isn't because the whole value changes. For example, if the `Array` constructor changes, that change is separate from the name of the function or whatever other args we pass to the function.

I don't know that `ArrayF _` is going to change a whole bunch as the code evolves, so it might be fine to do a top-level pattern match. I also don't know that `ArrayF _` is going to stay the same as the code evolves, so it seems safer to have it as a regular case statement.

We can definitely write it as a top-level pattern match if that's what you'd prefer.

src/Data/Expr.purs
joneshf-cn: Are you okay with doing that change separate from this PR? Ideally we can add a test suite around `print` so we don't have to think about whether code changes are altering behavior. The changes in this PR don't alter the way `print` currently works. I.e. On both `master` and this branch `print (array_ [int_ 1, int_ 2, int_ 3]) == "[1, 2, 3]"`.

src/Data/Expr.purs
joneshf-cn: Yeah, exactly. `Val _` and `BooleanF _` are phantom types. There's no value to pass to any function. So, there's nothing to do but to ignore it.

src/Data/Expr.purs
joneshf-cn: More than happy to move to whatever, but as with the `print` changes for arrays, would you be okay with doing these changes in a separate PR?

src/Data/Expr.purs
davezuch: > With these Haskell-based languages, top-level pattern matches get pretty noisy. It's harder to tell at a glance what is important and what isn't because the whole value changes. For example, if the Array constructor changes, that change is separate from the name of the function or whatever other args we pass to the function.

That's a solid argument for separating the function name from the pattern matching.

> We can definitely write it as a top-level pattern match if that's what you'd prefer.

I don't have a preference. I value conciseness, I also value consistency, and separation of concerns.

Consistency could mean always going for `case`, or it could be a little more nuanced, i.e. pattern matching at the top level when it's only one constructor, casing otherwise. As long as we're consistent. 

I do think when dealing with newtypes it's valid just to use `unwrap`. For example, what about `let` bindings or inline functions?

```purescript
_.teamId <<< unwrap <$> maybeTeam
```
I think is a lot nicer than getting at that value via casing. Of course, it's a different scenario than here, but it's still a question of where to draw the scope of this rule. I might be overthinking.

Anyway, open to having `case` be the standard, was just looking for clarification on why it's your default. 🙂

src/Data/Expr.purs
davezuch: Absolutely. I was actually thinking that this morning, that the `print` changes should be done separately. I like the idea of adding a test suite around `print`, since a type signature of `a -> String` doesn't have any guarantees about what the function of that string is

src/Data/Expr.purs
davezuch: Yup! Mostly thinking out loud, but also looking for feedback on those thoughts. Happy to move the discussion elsewhere though since it's not really relevant to this PR

src/Data/Expr.purs
joneshf-cn: Cool. What should we do here, top-level or case? Either one seems fine, and I'm hapy to change this code before merging.

davezuch: I'm down to keep this and nail down our rules around pattern matching later

src/Data/Expr.purs
davezuch: I am sort of feeling the pattern matching to be unnecessary here. We don't actually care about the value or the constructor, in which case I think it makes sense to just `foldMap _ _ = mempty`

PR #22, opened 2019-04-12
PS 2986 Make lynx usable from outside the package

joneshf-cn:
## What it does

Separates out the core component logic of lynx from all the hardcoded stuff we had. The hardcoded examples still work fine, but now we can use it in our internal projects.

## Review asks

* Did we get the boundaries correct? It seems like the component should only really care about expressions, loading the expression can be done by the parent. Is that right?


src/Lynx.purs
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+data Query a
+  = EvalForm (Page Expr) a
+  | UpdateValue Key (InputSource ExprType) a
+  | DropdownQuery Key (Dropdown.Message Query ExprType) a
+  | DateTimePickerQuery Key DateTimePicker.Message a
+  | TypeaheadSingleQuery  Key (Typeahead.Message Query Maybe ExprType) a
+
+type ParentInput = RemoteData EvalError (Page Expr)
```
davezuch: I'm not sure about this one. Makes sense for `RemoteData` to come from the parent, but `EvalError` comes from `Lynx`. Although I don't think I'd want to nest our data structure to something like `RemoteData e (Either EvalError (Page Expr))` 😅

What do you think about something like `EvalError e` being an open variant, so we could do something like:
```purescript
type ParentInput e = RemoteData (EvalError e) (Page Expr)
```

davezuch: Another option is to separate concerns and handle failure of `RemoteData` in the parent, and only load Lynx on success, passing only `Page Expr` down as input. This might make more sense, as different parents might want to handle failure differently.

davezuch:
Thanks for doing this @joneshf-cn! Was something that would need to be separated out soon anyway. 

I actually think we may need to go further soon, at least if we get the go-ahead for Lynx on the backend. In that case, we'd probably want to break out the `Data` modules into a core library, and then have different drivers for running the forms in different contexts. A Halogen driver, for running forms in a browser, which would be the `Lynx` module you broke out in this PR. As well as a simpler driver for evaluating forms on the server.

Open to other thoughts on this though.

src/Lynx.purs
joneshf-cn: Good point. `EvalError` is completely an implementation detail of lynx (that will hopefully go away sooner rather than later).

I think making input just `Page Expr` seems the most reasonable.

joneshf-cn:
I might be misunderstanding what the server would need to do, but it seems like `Lynx.Data.Form.eval` is that simpler driver. I guess it'll become clearer once we start implementing stuff, so I don't need a full rundown of the differences.

src/Lynx.purs
```diff
@@ -0,0 +1,277 @@
...
+import Ocelot.Block.Toggle (toggle) as Toggle
+import Ocelot.Component.DateTimePicker as DateTimePicker
+import Ocelot.Component.Dropdown as Dropdown
+import Ocelot.Component.Dropdown.Render as Dropdown.Render
+import Ocelot.Component.Typeahead as Typeahead
+import Ocelot.Data.Currency (parseCentsFromDollarStr)
+import Ocelot.HTML.Properties (css)
+
+type State =
+  { form :: RemoteData EvalError { expr :: Page Expr, evaled :: Page ExprType }
```
davezuch: I don't think we need `RemoteData` here anymore, right? That's now a concern of the parent. I think this can just be
```purescript
{ form :: Either EvalError { expr :: Page Expr, evaled :: Page ExprType }
...
}
```
Or do you think it makes sense to split these out now to something like
```purescript
{ expr :: Page Expr
, evaled :: Either EvalError (Page ExprType)
...
}

davezuch:
>  it seems like Lynx.Data.Form.eval is that simpler driver. I guess it'll become clearer once we start implementing stuff

Yeah, I think you're right on both counts.

src/Lynx.purs
joneshf-cn: Oh nice!

So what's the difference between those two? One is saying that you either have some error from trying to evaluate the page or you have both the page and its evaluated form. The other is saying that you always have the page and either some error from trying to evaluate the page or its evaluated form. Which one is the correct semantics?

joneshf-cn: I think I'm going to drop the two cases of `RemoteData` for now, and we can think about changing the semantics later.

davezuch: > So what's the difference between those two? One is saying that you either have some error from trying to evaluate the page or you have both the page and its evaluated form. The other is saying that you always have the page and either some error from trying to evaluate the page or its evaluated form. Which one is the correct semantics?

 I think the latter, no? Because we do always have the page, just sometimes we'll run into errors when evaluating it. Plus it allows us to recover the unevaluated page upon failure, if we for some reason needed to.

davezuch (approved):

src/Lynx.purs
joneshf-cn: MAkes senes. I'll throw together a PR for it.

PR #23, opened 2019-04-15
Change the way we deal with state

joneshf-cn:
Re: https://github.com/citizennet/purescript-lynx/pull/22#discussion_r275532240

## What it does

In the above discussion, we realized that we'd been modeling the state incorrectly. The changes here fix it so it's more accurate to our mental model.

## Review asks

Nothing in particular.


davezuch (approved):
Cleans things up a little, I think this was the right call 👌

PR #24, opened 2019-04-16
PS-2988 Allow a page to contain tabs

joneshf-cn:
## What it does

Allows us to have tabs in our pages. We're not currently displaying the tabs. That will come when we work on PS-2989.

## Review asks

Might be useful to look at the diff without whitespace. Much of the changes are re-indenting.


davezuch (approved):
giphy-1

PR #25, opened 2019-04-16
Use non-empty arrays

joneshf-cn:
## What it does

As discussed in slack, we don't really have a case for empty arrays. This shouldn't change much from the outside, except disallowing empty arrays in the JSON blobs.

## Review asks

* We couldn't use `NonEmptyArray _` and had to use `NonEmpty Array _` instead. There's no `DecodeJson`/`EncodeJson` instances for `NonEmptyArray _`. `NonEmpty _ _` is known to be less performant than a direct encoding, but it seems like we're not going to hit the numbers that would show that (we're probably in the ten's of elements). Are we okay with this choice or should we newtype `NonEmptyArray _` and write some instances?


davezuch (approved):
> There's no DecodeJson/EncodeJson instances for NonEmptyArray _. NonEmpty _ _ is known to be less performant than a direct encoding, but it seems like we're not going to hit the numbers that would show that (we're probably in the ten's of elements). Are we okay with this choice or should we newtype NonEmptyArray _ and write some instances?

I think we can go with `NonEmpty Array _` for now. The long-term answer might be a PR to the argonaut lib to add instances for `NonEmptyArray _`, but I agree that the performance difference may not even be noticeable

PR #26, opened 2019-04-18
PS 2989 Render tabs

joneshf-cn:
## What it does

We can now display the tabs we have!

## Review asks

* Should we do something different with the UI here?

    <details>
        <summary>Tab UI</summary>
        Screen Shot 2019-04-17 at 7 16 40 PM
    </details>

    It looks a little off because the tabs have no left margin/padding. On the site, there's some margin/padding on the container.
* I might have some contextual questions. If so, I'll make review for them.


src/Data/Form.purs
```diff
@@ -223,6 +223,15 @@ noErrors (Errors e) = Data.Set.isEmpty e
 errorsToArray :: ∀ e. Ord e => Errors e -> Array e
 errorsToArray (Errors e) = toUnfoldable e
 
+errors :: forall a. Field a -> Errors ValidationError
```
joneshf-cn, line 226: This function exists because the `errors` field is within each `Input _`. Should we move this up to the top level, or should we leave it where it is as we plan to have errors specific to each case?

This is more a general question that we should address in a separate PR if there's something to address.

src/Data/Form.purs
```diff
@@ -642,6 +664,19 @@ mvpObjective =
       ]
     )
 
+mvpSocialAccount :: Field Expr
```
joneshf-cn, line 667: I threw together this as a toggle because toggles are simple. It's not intended to be anything more than a simple mock up to allow us a second tab. Happy to change it to something else if we'd like.

joneshf-cn (commented):
Guess I did have a couple things to say :).

src/Data/Form.purs
davezuch: Hmm, yeah I guess we need to circle back and figure out how we want to model errors. That might be something to block out sprint work for so that we actually have time to focus on it.

davezuch (approved):
> It looks a little off because the tabs have no left margin/padding. On the site, there's some margin/padding on the container.

😂

I can add my own PR to fix the stylings a bit, but I think this is fine for now.

PR #27, opened 2019-04-18
Containerize

davezuch:
## What it does

Adds configs and documentation for deploying Lynx into a local k8s cluster.

## Review asks

Is this something that belongs in this repo, or should we have a separate repo where we manage deployment configs for our various libraries? I tried to decouple the configs as much as possible from anything CitizenNet-specific, so that anyone who has a local k8s cluster can deploy this project to it. I think I succeeded, but I won't really know until someone tries. But also, is that even something worth worrying about at the moment?

## How should the reviewer test manually?

Would be useful if someone else pulled this down and verified that it works for them or not. That might help to surface any issues I haven't accounted for.


Makefile
```diff
@@ -74,7 +74,7 @@ RTS_ARGS ?=
 .DEFAULT_GOAL := dist/main.js
 
 $(BOWER_COMPONENTS): bower.json $(NODE_MODULES)
-	npx bower install
+	npx bower install --allow-root
```
joneshf-cn: Can we make this an argument we pass in (like we do for `RTS_ARGS`)? If you're not using this makefile within a docker container, it shouldn't be necessary to use this flag.

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM node:stretch
+
+# Global install yarn package manager
```
joneshf-cn, line 3: I'm more than down to use `yarn` for everything. Let's move the `Makefile` over to that as well. We can do that in a separate PR if you'd like though.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# Global install purescript, pulp, and bower
```
joneshf-cn: Is it possible to have project local installs of these programs, or is there something about being in a container that makes this not so easy? If it's possible, can we let the `package.json` file be the source of truth for project dependencies? I'm worried about getting drift between the dependencies in this `Dockerfile` and the `package.json`.

Makefile
```diff
@@ -129,3 +129,7 @@ test: dist/main.js $(BUILD)/test.out
 .PHONY: watch
 watch: $(BOWER_COMPONENTS) $(NODE_MODULES)
 	npx watch-exec --command 'make dist/main.js' --watch $(SRC)
+
+.PHONY: serve
+serve:
```
joneshf-cn: I think if we specify the `dist/main.js` file as a dependency:
```Makefile
serve: dist/main.js
```
It'll always ensure that `dist/main.js` is up to date before serving the files (including installing dependencies).

joneshf-cn: If we have that, maybe we can change the `watch-exec` command to `make serve`? I'll check on this and report back if you haven't tried that already by the end of the PR.

.k8s/templates/_helpers.tpl
```diff
@@ -0,0 +1,32 @@
+{{/* vim: set filetype=mustache: */}}
```
crcornwell, line 1: I think some of this could be simplified or removed. For instance, there are handlers in here for supporting a full name override flag. Also, we don't really need to append `-standard-web` to the deployment name.

.k8s/templates/deployment.yaml
```diff
@@ -0,0 +1,71 @@
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  name: {{ template "standard-web.fullname" . }}
```
crcornwell: Some of these we might just be able to hard code, like `standard-web.fullname`, especially if you're going to implement one of these in each library. That may even allow you to get rid of the `_helpers.tpl` entirely. I'm not sure about that though.

If you want to break this out and just have one set of these configs that you can use across ocelot and lynx, then this generalization will be important.

crcornwell (approved):

Makefile
joneshf-cn: Looks like this works! That means we can have just the one container running `make watch` and everything should work.

Makefile
```diff
@@ -129,3 +129,7 @@ test: dist/main.js $(BUILD)/test.out
...
+	http-server ./dist -p 80
```
joneshf-cn: Can we also take the port as an argument (defaulting to `80`)? If someone is running `make serve` outside of the k8s workflow (not as root), it's likely not going to work.

dist/index.html
```diff
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8">
     <title>CitizenNet UI Guide</title>
+    <base href="/lynx/" />
```
joneshf-cn: Looks like this breaks the page unless we're serving this directory as `/lynx`. Can we drop this in favor of fixing the ingress to point to `/lynx/`? That should make this page work both in our k8s workflow and for normal HTML stuff.

.k8s/README.md
```diff
@@ -0,0 +1,25 @@
+# Deploying into local Kubernetes cluster
```
joneshf-cn, line 1: 😻 Thanks for writing this up!

joneshf-cn (commented):
This is great! I've a couple of things that we should address/discuss, but otherwise looks good to me!

joneshf-cn:
Sorry, I thought I'd replied to this last night

---

> Is this something that belongs in this repo, or should we have a separate repo where we manage deployment configs for our various libraries?

I think this is the right place for it. The configuration here is specific to lynx. It seems to make sense that it live close to it.

> I tried to decouple the configs as much as possible from anything CitizenNet-specific, so that anyone who has a local k8s cluster can deploy this project to it. I think I succeeded, but I won't really know until someone tries. But also, is that even something worth worrying about at the moment?

I tried deploying this on my local laptop, but I can't figure out what my hostname is supposed to be. `$(hostname)` gives me back some gibberish (I think it's conde gibberish), `localhost` doesn't seem to connect, and I can't feed it an IP address. Maybe this is something inherent to Ingress and not specific to this project though?

I don't think it's something huge to worry about, as I think anyone not using our internal setup wouldn't bother setting up k8s and use `helm` just to work on this project.

.k8s/templates/_helpers.tpl
davezuch: Better?

Dockerfile
davezuch: Down! Agreed that it might be appropriate for a separate PR though

.k8s/templates/deployment.yaml
```diff
@@ -0,0 +1,71 @@
...
+          image: "lynx-local:latest"
+          command: ["make", "serve", "PORT=80"]
+          volumeMounts:
+            - name: src
+              mountPath: /usr/src/app
+          imagePullPolicy: IfNotPresent
+          env: []
+          ports:
+            - name: http
+              containerPort: 80
```
davezuch, line 36: All the k8s ports are hardcoded, but unlike the `Makefile`, I'm not sure it's worth paramaterizing these for now

Dockerfile
davezuch: I removed everything except `http-server`, as that needs to run from `root` for port `80` to work

davezuch:
> I tried deploying this on my local laptop, but I can't figure out what my hostname is supposed to be. `$(hostname)` gives me back some gibberish (I think it's conde gibberish)

Ah, yeah I guess that bit assumes you've followed @crcornwell's Minikube instructions, I'll add them to the readme, but if you do this I think you'll get it to work:

```bash
hostname [your-initials].aws.citizennet.com
hostname > /etc/hostname
```

Dockerfile
joneshf-cn: For audit, we also spoke in slack a bit about this. We might be able to switch the port to a non-root port and port forward while in a container. Punting on doing it ini this PR, though.

joneshf-cn: Oh wait, the `make serve` recipe doesn't work unless you're in a container. I think we can move `http-server` to a project local install and `make serve` should work both within a container if `root` is running it and outside of a container if a normal user is running it.

Makefile
joneshf-cn: For audit, we talked abotu this in slack. It sometimes ran into issues. Punting on figuring it out in this PR.

joneshf-cn (approved):
Thanks for updating things!

There's something with `make serve` that I commented on, but I'm more than happy to deal with that in a separate PR (as I think I'm the only person it affects right now).

Dockerfile
davezuch: I made the default port `8080` so that `make serve` will work from outside a container (`make serve PORT=80` won't though)

joneshf-cn:
> Ah, yeah I guess that bit assumes you've followed `@crcornwell`'s Minikube instructions, I'll add them to the readme, but if you do this I think you'll get it to work:

Sorry for being confusing. I was trying to run this on my local laptop. I don't really want to mess with `/etc/hostname` just to get this to work on it. I'm not sure what needs to be different in that world, but I don't think we really need to worry about it in this PR.

PR #28, opened 2019-04-19
Generate fewer test cases for larger data

joneshf-cn:
## What it does

As mentioned in standup, the test are super slow (10 minutes on my personal laptop). The slowness is a two-parter: `quickCheck` generates 100 examples, and we don't control the `Arbitrary` instances for records. The simple fix was moving over to `quickCheck'` and passing a lower number of tests. These roundtrips are mostly sanity checks and testing the entire page doesn't need to check so many examples (it's a product of sums and we check the sums with many examples, so a few tests will weed out problems).

## Review asks

Nothing in particular.


davezuch (approved):
Glad this was so easy!

joneshf-cn:
Thanks for the suggestion!

PR #29, opened 2019-04-26
PS-2989: Improve form styles

davezuch:
## What it does

Makes the form a little more presentable for demo tomorrow. It's not pretty, but it's prettier.


joneshf-cn (approved):
Looks great!

Thanks for doing this!

PR #30, opened 2019-04-26
Only evaluate expression when the fragment changes

joneshf-cn:
## What it does

Before, we were evaluating the new expression all the time. That meant when we changed tabs, we'd lose all of the state we had in our page. This isn't the behavior we wanted. We want the state between tabs to stay. This change allows us to keep state between tabs.


davezuch (approved):

davezuch:
Nice catch!

PR #31, opened 2019-05-06
Joneshf cn/update dependencies

joneshf-cn:
## What it does

Similar to https://github.com/citizennet/purescript-ocelot/pull/88, this locks down the dependencies in hope that version bounds won't get unmanageable for consumers. It also drops some unused dependencies that have been hanging around.


davezuch (approved):
Looks great, thanks for updating here!

PR #32, opened 2019-05-07
PS-2986 More prep for consumption

joneshf-cn:
## What it does

Looks like we missed a few more things when trying to get `lynx` ready to be consumed by our internal project. We fix the module namespace so things don't conflict (`Main`) and are a bit easier to understand (`Lynx.Expr`/`Lynx.Form` vs. `Lynx.Data.Expr`/`Lynx.Data.Form`). We also rename the `bower.json` so it can be installed with a name we recognize.

## Review asks

Does everything look okay here?


davezuch (approved):
Looks a lot better without all the `.Data`'s actually!

PR #33, opened 2019-05-29
Be a bit more correct with dependencies

joneshf-cn:
## What it does

Re: https://github.com/citizennet/orders/pull/44#discussion_r288305858. We fix our dependencies to be what we actually need in the `src` directory.


davezuch (approved):
Thanks for updating here @joneshf-cn!

PR #34, opened 2019-05-30
PS-3015 Add collections of expressions

joneshf-cn:
## What it does

Putting this up as a draft PR, since I'm not sure this is the correct approach yet.

## Review asks

There are some things to figure out as far as how to de/serialize data. I'll make a review so it's contextual.

### DIFF
--- a/src/Lynx/Expr.purs
+++ b/src/Lynx/Expr.purs
@@ -4,6 +4,7 @@ import Prelude
 
 import Control.Alt ((<|>))
 import Data.Argonaut (class DecodeJson, class EncodeJson, Json, decodeJson, encodeJson, jsonEmptyObject, stringify, (.:), (:=), (~>))
+import Data.Array as Data.Array
 import Data.BigInt as Data.BigInt
 import Data.DateTime (DateTime)
 import Data.Either (Either(..))
@@ -516,6 +517,7 @@ data Expr
   | Equal Expr Expr
   | Print Expr
   | Lookup Key Expr
+  | Collection (Array Expr)
 
 derive instance eqExpr :: Eq Expr
 
@@ -571,6 +573,7 @@ data ExprA
   | EqualA ExprA ExprA String String
   | PrintA ExprA String String
   | LookupA Key ExprA String
+  | CollectionA (Array ExprA) String String
 
 instance encodeJsonExprA :: EncodeJson ExprA where
   encodeJson =
@@ -584,6 +587,8 @@ instance encodeJsonExprA :: EncodeJson ExprA where
         "params" := [ x ] ~> encodeHelper "Print" i o
       LookupA x y o ->
         "params" := [ encodeJson x, encodeJson y ] ~> encodeHelper "Lookup" "Void" o
+      CollectionA x i o ->
+        "params" := encodeJson x ~> encodeHelper "Collection" i o
 
 toExprA :: Expr -> ExprA
 toExprA = case _ of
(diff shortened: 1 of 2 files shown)


src/Lynx/Expr.purs
```diff
@@ -592,6 +597,11 @@ toExprA = case _ of
   Equal x y -> EqualA (toExprA x) (toExprA y) (reflectIn x) (reflectOut x)
   Print x -> PrintA (toExprA x) (reflectIn x) "String"
   Lookup x y -> LookupA x (toExprA y) (reflectOut y)
+  Collection x ->
+    CollectionA
+      (map toExprA x)
+      (maybe "TODO" reflectIn $ Data.Array.head x)
```
joneshf-cn, line 603: How do we deal with the types here? Does it matter? Maybe we need to require at least one element? Maybe we don't need to talk about individual types too deeply? Maybe we're thinking at the wrong level?

citizengabe: Do you mean the type of `x`?

joneshf-cn: I mean the whole collection. The `ExprA` data type is supposed to be `Expr` annotated with the input and output types of the expression. E.g. the input type of `Equal` is whatever the input types of the sub `Expr`s are, and the output type of `Equal` is always `Boolean`.

For the idea of collections, what is its input type, and what is its output type? I would think we're dealing with homogeneous collections of things, so the input and output type would be related to the items in the collection. But if we don't have any element, where do we get that type information?

davezuch: Is `Collection` in the right place? That is, should it be a `Expr` as opposed to an `ExprType`? Which, we already have `Array` for. It's been a while since I've really dealt with this module, but I'm having trouble figuring out what we'd do with this `Collection`.

joneshf-cn: `Expr` is used to define things in `Lynx.Form`. `ExprType` is the evaluated value of the `Expr`. If we put `Collection` in `ExprType`, we cannot evaluate it (it's already evaluated) and it effectively becomes static. I assumed we wanted collections of expressions to be dynamic. Is that not correct?

Looking at what we do for `Array`, we say the type is `"Array"`, we don't look underneath at any elements. It seems like we should do similar for `Collection` and say its type is `"Collection"`. That seem right?

davezuch: > If we put `Collection` in `ExprType`, we cannot evaluate it (it's already evaluated) and it effectively becomes static. I assumed we wanted collections of expressions to be dynamic. Is that not correct?

My understanding of why we need `Collection`, is for the values of "Array" input types. Right now values are always static. I'm not sure what it would mean if a user could set the value of a field to an expression.

I.e., right now [`Lynx.Expr.Form.setValue` takes `InputSource ExprType` as an argument](https://github.com/citizennet/purescript-lynx/blob/master/src/Lynx/Form.purs#L456). If we were to take `InputSource Expr` instead, we'd have to allow users to build expressions in our UI. Something we will need to figure out for the form builder, actually, but I don't think that's something we are trying to support for regular Lynx forms.

joneshf-cn: Is there anything to do on the `Expr`/`ExprType` side? If we don't need to have arrays of things at the `Expr` level, it seems like there's no change required. I'm more than happy to close this PR.

davezuch: Hmm, yeah I'm not sure there is. I say perhaps we start with adding arrays/collections to our layout types and see how far we can get with that.

joneshf-cn: Sounds good. Thanks for the review you two!

joneshf-cn:
Closing per the above discussion

PR #35, opened 2019-06-05
WIP: PS-3013 Expand Section to include single and a collection

citizengabe:
## What it does

Please describe the purpose of the pull request, including any background context and links to related issues. If this introduces a large change to the project, please describe what you have implemented.

This PR implements https://citizennet.atlassian.net/browse/PS-3013

It promotes the `Section` alias to a data type w/ two constructors, `Single` and `Collection` (the names are not the best). `Collection` differs from `Single` only in an added `key` property.

## Review asks

Please describe any questions or concerns this pull request introduces, and notable pros and cons associated.

This is really a jumping off point for discussion of the requirements for PS-3013. What are the requirements for "The Array"?

@davezuch I'd appreciate your input on this.


src/Lynx/Form.purs
```diff
@@ -40,7 +40,38 @@ type Page f = Record (LayoutRows (Tab f) ())
...
+
+-- TODO: Naming is hard
+data Section f
+  = Single (SectionRecord f ())
+  | Collection (SectionRecord f (key :: String))
+
+section :: forall f a. (forall r. SectionRecord f r -> a) -> Section f -> a
+section f = case _ of
+  Single r -> f r
+  Collection r -> f r
```
davezuch, line 53: Love this helper function.

src/Lynx/Form.purs
```diff
@@ -40,7 +40,38 @@ type Page f = Record (LayoutRows (Tab f) ())
 
 type Tab f = Record (LayoutRows (Section f) (link :: String))
 
-type Section f = Record (LayoutRows (Field f) ())
+type SectionRecord f r = Record (LayoutRows (Field f) r)
+
+-- TODO: Naming is hard
+data Section f
+  = Single (SectionRecord f ())
+  | Collection (SectionRecord f (key :: String))
```
davezuch, line 48: So, something's been bothering me about this type. 

We have `Single`, which is used solely as a layout construct, for grouping together a collection (heh) of related fields in the UI, but has no bearing on the data model. I.e., a form with a single `Single` containing the fields `name`, `age`, `email`, will produce a model with the shape: `{ name, age, email }`.

Then we have `Collection`, which similarly groups together related fields in the layout, but in the UI you can repeat this collection 0 or more times, and that has a direct bearing on the data model, meaning each of those collections represents an array value for the `key` defined on the `Collection`. I.e., a form with a single `Collection` with `key: "profiles"` containing the fields `name`, `age`, `email`, will produce a model with the shape: `{ profiles: [{ name, age, email }] }`

I think what's bothering me, is the grouping of these two types together, and their naming, suggests that `Single` should also have a key, and the resulting model shape should be more akin to: `{ profile: { name, age, email } }`

That's not our intent, but it does raise the question: is that something we want to support? A collection of fields grouped under a single key in our model, not represented as an array but as a record/map. Probably (though not right now, that would be scope creep).

So, maybe the thing to do is come up with better naming (famously one of the two hardest things about programming) and move on. But still, mixing the semantics of our types (purely layout vs. layout+data model) does bother me. Ignoring that, perhaps we go with something like:

```purescript
data Segment f -- a segment of our form
  = Card (SegmentRecord f ()) -- our segment is just a UI card
  | FormList (SegmentRecord f (key :: String)) -- an array on our model, displayed as a series of UI cards for each array value
  | FormCollection (SegmentRecord f (key :: String) -- a collection of values on our model grouped under a single key, displayed as a single UI card
```

My problem with this is it limits us in a number of ways. What if we want to group a large number of fields together under one key, but want them to be rendered in more than one UI card? What if we want more than one UI card to render all the fields in an array on our model?

The former could be solved if we updated `LayoutRows` with a `(key :: Maybe String)`. Not a huge fan of propagating maybe keys through all of our layout types, but it would give us what we need there. The second problem seems harder to solve. Maybe if we extended `Tab` and `Section` with a shape akin to:

```purescript
type Tab f = Record (LayoutRows (Section f /\ ArrayOf Section f) (link :: String))

type Section f = Record (LayoutRows (Field f /\ ArrayOf Field f) ())

type ArrayOf a f = Record (key :: String, contents :: a f)

--Not saying we need to use `Coproduct` over standard ADT's, just felt like an easier way to illustrate the idea
```

I think that would work, but I don't know how I feel about it. Am I worrying too much about this now? Is this something we can get kind of wrong for now and fix later? Open to better idea's here @joneshf-cn and @citizengabe.

src/Lynx/Form.purs
```diff
@@ -731,16 +768,16 @@ testPage =
...
-  , contents:
-    Data.NonEmpty.NonEmpty
-      firstName
-      [ lastName
-      , active
-      , food
-      , money
-      ]
-  }
+  Single { name: "Name"
```
davezuch, line 771: Can we put the record on a new line, indented one level past `Single`?

davezuch (commented):
Overall this looks great! I left one giant comment describing my concerns with the approach we landed on. Not opposed to going with this and addressing those concerns later, if that turns out to be the consensus.

src/Lynx/Form.purs
citizengabe: > What if we want to group a large number of fields together under one key, but want them to be rendered in more than one UI card?

Do you mean having a segment defined using a single key but have it rendered in multiple cards?

> What if we want more than one UI card to render all the fields in an array on our model?

This feels related to our discussion about list views. We're going to want to be very flexible about how we display our model. We may potentially want to display data from one property in different components in different formats or combine data from different properties into a single component.

> The former could be solved if we updated LayoutRows with a (key :: Maybe String). Not a huge fan of propagating maybe keys through all of our layout types, but it would give us what we need there.

Couldn't we just say the empty string represents no `key` and deal with that render time?

```haskell
type Tab f = Record (LayoutRows (Section f /\ ArrayOf Section f) (link :: String))
...
```
Isn't `Section f /\ ArrayOf (Section f)` isomorphic to `ArrayOf (Section f)`?

I hope my responses makes sense.

I think I need to spend some time reading through `lynx` _and_ `orders` to get a real grasp of what our intent is. For instance, I don't understand how we serialize the data model defined by the user in the template. Are campaign models totally ad-hoc or does that take manual on the part of the developer?

joneshf-cn: > We have `Single`, which is used solely as a layout construct, for grouping together a collection (heh) of related fields in the UI, but has no bearing on the data model. I.e., a form with a single `Single` containing the fields `name`, `age`, `email`, will produce a model with the shape: `{ name, age, email }`.
> 
> Then we have `Collection`, which similarly groups together related fields in the layout, but in the UI you can repeat this collection 0 or more times, and that has a direct bearing on the data model, meaning each of those collections represents an array value for the `key` defined on the `Collection`. I.e., a form with a single `Collection` with `key: "profiles"` containing the fields `name`, `age`, `email`, will produce a model with the shape: `{ profiles: [{ name, age, email }] }`

When you say model, what do you mean? If we're talking about the JSON, I think the JSON that's created has a `type` key that discerns which case we're on, much like we do for `Input` and `InputSource`. I.e. the JSON would be like: `{ "contents": …, "name": …, "type": "Single" }` and `{ "contents": …, "key": …, "name": …, "type": "Collection" }`, respectively.

Does that change your concerns? If not JSON, which model are we talking about?

> maybe the thing to do is come up with better naming … and move on.

I'm kind of of a few minds here:
* We don't know everything we're building.
* We individually use different terminology (me, you, Spencer, etc.). We were talking about "arrays" in sprint planning, I named it "collection" in #34, you've mentioned "segments." I feel like we're adding excess confusion by not all being on the same page.
* We use different terminology over time. Once we're on similar pages, we change that over time. That's fine, that's kind of how tech debt works. But it's not readily apparent when we're changing terms or which terms we're changing.

The latter two points make me want to commit a glossary in the repo and adhere to those terms dogmatically throughout the company. If a term is being renamed, cool, let's rename it. But, do it throughout the entirety of the repo and use it going forward in all conversations.

The first point makes me want to take an even more iterative approach rather than a big upfront design approach. You've done a huge amount of work to flesh out requirements and document a bunch of stuff. That has helped immensely to get where we are. It's seems that now we're getting to some things that we simply don't know how they should be.

I think getting stymied on some of these unknowns belies the use of PureScript as the implementation language. Even if we make a mistake and create the wrong thing, we ought to be able to fix that with lower cost later on than if we were using a different language. It seems like we can use PS to our advantage, do the minimum we can without worrying too much about whether we're shooting ourselves in the foot, and make adjustments when our past decisions cause problems.

davezuch: > Do you mean having a segment defined using a single key but have it rendered in multiple cards?

Exactly.

> Couldn't we just say the empty string represents no `key` and deal with that render time?

Could, but a `Maybe String` is more explicit/clear. Perhaps our keys should actually be `NonEmptyString` 🤔 

> Isn't `Section f /\ ArrayOf (Section f)` isomorphic to `ArrayOf (Section f)`?

`ArrayOf` would be our way of wrapping the contents of our `Section f` in an array (and the key would be required). E.g. if we have a `sectionA` with `name`, `age`, and `email` fields, and `key: Nothing`, it'd be represented on the model as: 
```
{ name, age, email }
```
If we have `sectionB` with `name`, `age`, `email`, and `key: Just "profile"`, it'd be represented on the model as:
```
{ profile: { name, age, email } }
```

If we wrap `sectionA` in `ArrayOf` with `key: "users"`, we end up with: 
```
{ users: [{ name, age, email }] }
``` 

If we wrap `sectionB` in the same, we end up with:
```
{ users: [{ profile: { name, age, email } }] }
```

> For instance, I don't understand how we serialize the data model defined by the user in the template.

Yeah, that hasn't been implemented yet.

> Are campaign models totally ad-hoc or does that take manual on the part of the developer?

Not sure I understand this question. A piece of this whole system that has yet to be implemented, or even really designed, is the mapping functions that define how to go from a custom Lynx form, to some other shape, be it a FB campaign or a different Lynx form.

citizengabe: >> Are campaign models totally ad-hoc or does that take manual on the part of the developer?

> Not sure I understand this question. A piece of this whole system that has yet to be implemented, or even really designed, is the mapping functions that define how to go from a custom Lynx form, to some other shape, be it a FB campaign or a different Lynx form.

Sorry, I missed an "effort" in there, "...does that take manual effort...".

I think you answered my question though. We have yet to define how a Lynx form is transformed into a campaign.

davezuch: > When you say model, what do you mean? If we're talking about the JSON, I think the JSON that's created has a `type` key that discerns which case we're on, much like we do for `Input` and `InputSource`. I.e. the JSON would be like: `{ "contents": …, "name": …, "type": "Single" }` and `{ "contents": …, "key": …, "name": …, "type": "Collection" }`, respectively.

Oh man, this comment made me realize there was a whole side to this problem that I was overlooking. So when I've been talking about the model, I've been thinking about the `values :: Map Key ExprType` on our `Lynx` state. So saying we need to support arrays on our model means we need an `ExprType` value for arrays (we have arrays) that is basically also `Map Key ExprType`. Does that mean adding `Map Key ExprType` to our `ExprType`?

To further complicate things (this is the part I hadn't thought of), we are coupling our values to our form definition (`value :: InputSource f` on our `Input`s), which, becomes a problem when we introduce arrays. I don't know that we can continue doing that. Maybe my [`ArrayOf`](https://github.com/citizennet/purescript-lynx/pull/35#discussion_r291686389) row get's updated with `SharedRows`, and the `value` ends up being an array of whatever its `contents` is. Not sure I like that. May be cleaner to separate our form definition from our values, though I did like the benefits it bought us.

> * We individually use different terminology (me, you, Spencer, etc.). We were talking about "arrays" in sprint planning, I named it "collection" in #34, you've mentioned "segments." I feel like we're adding excess confusion by not all being on the same page.

You're completely right (at least we're on the same page about this). I think it'd be worthwhile to document an agreed upon set of terminology so that we're all on the same page, and update it as terminology as added or changed.

joneshf-cn: > when I've been talking about the model, I've been thinking about the `values :: Map Key ExprType` on our `Lynx` state.

Gotcha, thanks!

> So saying we need to support arrays on our model means we need an `ExprType` value for arrays (we have arrays) that is basically also `Map Key ExprType`. Does that mean adding `Map Key ExprType` to our `ExprType`?

Why is that? We're not altering the structure of our `Page _`, so each `value` in a `Field` shouldn't need to have multiple `ExprType`s. The thing that seems like it would have to change is the lookup logic.

If we have:
```PureScript
{ name: "page"
, contents:
  [ { name: "tab"
    , link: "tab"
    , contents:
      [ { name: "collection"
        , key: "collection"
        , contents:
          [ { name: "field1"
            , key: "field1"
            , input:
              Toggle
                { value: UserInput (val_ $ boolean_ true)
                …
                }
            …
            }
          ]
        }
      ]
    }
  ]
}
```

When we evaluate it, it ends up looking like:
```PureScript
{ name: "page"
, contents:
  [ { name: "tab"
    , link: "tab"
    , contents:
      [ { name: "collection"
        , key: "collection"
        , contents:
          [ { name: "field1"
            , key: "field1"
            , input:
              Toggle
                { value: UserInput (boolean_ true)
                …
                }
            …
            }
          ]
        }
      ]
    }
  ]
}
```
The structure of the `Page _` hasn't changed. Finding that value does require knowing the `Key` of the array/collection and the `Key` of the `Field`. We ought to be able to change our model to reflect that. We could use `Map (EIther Key { collection :: Key, field :: Key }) ExprType`–or something equivalent.

> To further complicate things (this is the part I hadn't thought of), we are coupling our values to our form definition (`value :: InputSource f` on our `Input`s), which, becomes a problem when we introduce arrays. I don't know that we can continue doing that.

What is the problem? It seems like everything continues to work like normal.

davezuch: I think some part of the requirements is being lost here. I'll try to walk through an example. Let's say we want a form that looks like this:

IMG_4660

This form represents an object with a collection of Users. The person filling out the form should be able to add any arbitrary number of Users to our form object and save it. Using what we have now, we would construct this form as:

```purescript
{ name: "Users Form"
, contents:
  [ { name: "Users"
    , link: "users"
    , contents:
      [ Collection
        { name: "User"
        , key: "users"
        , contents:
          [ { name: "Name"
            , key: "name"
            , input: Text
              { value: NotSet
              ...
              }
            }
          , { name: "Email"
            , key: "email"
            , input: Text
              { value: NotSet
              ...
              }
            }
          ]
        }
      ]
    }
  ]
}
```
Coupling our value to our form structure, this only renders a single User card. Filling in our first User card, we might get this:

```purescript
{ name: "Users Form"
, contents:
  [ { name: "Users"
    , link: "users"
    , contents:
      [ Collection
        { name: "User"
        , key: "users"
        , contents:
          [ { name: "Name"
            , key: "name"
            , input: Text
              { value: UserInput (string_ "Hardy Jones")
              ...
              }
            }
          , { name: "Email"
            , key: "email"
            , input: Text
              { value: UserInput (string_ "hardy@citizennet.com")
              ...
              }
            }
          ]
        }
      ]
    }
  ]
}
```
We need `n` number of User cards. What happens when a user clicks `Add` to add a new User? How could that fit in to what we have here? What happens when they `x` all the User cards, how do we represent that state in our data structure without losing our form definition for Users?

I think the only way we can continue to couple our values to our data structure is to copy the concept of a template into our structure. At which point, maybe we need child forms after all. Something like:
```purescript
{ name: "Users Form"
, contents:
  [ { name: "Users"
    , link: "users"
    , contents:
      [ Collection
        { name: "User"
        , key: "users"
        , template: -- `contents` replaced by `template`
          [ { name: "Name"
            , key: "name"
            , input: Text
              { value: NotSet
              ...
              }
            }
          , { name: "Email"
            , key: "email"
            , input: Text
              { value: NotSet
              ...
              }
            }
          ]
        , value: NotSet -- `value` on our `Collection`
        }
      ]
    }
  ]
}
```
So the state of the mock form I drew above, would look like this:
```purescript
{ name: "Users Form"
, contents:
  [ { name: "Users"
    , link: "users"
    , contents:
      [ Collection
        { name: "User"
        , key: "users"
        , template: -- template for our `Collection`
          [ { name: "Name"
            , key: "name"
            , input: Text
              { value: NotSet
              ...
              }
            }
          , { name: "Email"
            , key: "email"
            , input: Text
              { value: NotSet
              ...
              }
            }
          ]
          -- each value in our array is a copy of the template
          -- that can be filled in by the user
        , value:
          UserInput
          ( array_
            [ [ { name: "Name"
                , key: "name"
                , input: Text
                  { value: UserInput (string_ "Hardy Jones")
                  ...
                  }
                }
              , { name: "Email"
                , key: "email"
                , input: Text
                  { value: UserInput (string_ "hardy@citizennet.com")
                  ...
                  }
                }
              ]
            , [ { name: "Name"
                , key: "name"
                , input: Text
                  { value: UserInput (string_ "Gabe")
                  ...
                  }
                }
              , { name: "Email"
                , key: "email"
                , input: Text
                  { value: NotSet
                  ...
                  }
                }
              ]
            ]
          )
        }
      ]
    }
  ]
}
```
Obviously it's not perfect, putting an array of `Field` inside `array_` being an obvious problem, but I hope the idea is a bit clearer.

joneshf-cn: Thanks for explaining what you're thinking!

> What happens when a user clicks `Add` to add a new User?  How could that fit in to what we have here?

It seems like the `Page _` should be altered. In particular, we should insert another element into `contents` of a collection.

> What happens when they `x` all the User cards, how do we represent that state in our data structure without losing our form definition for Users?

If they remove all the elements of a collections, we should remove all of the elements in the `contents`. The schema for what each element of `contents` is shouldn't come from the actual elements of `contents`. That method complicates things in the way you're explaining here. The schema should be known outside of the `contents`.

The concern you're explaining here is why I suggested that we work on [PS-3018](https://citizennet.atlassian.net/browse/PS-3018?atlOrigin=eyJpIjoiNTBjZGVlMzFhMjcyNDEzZTk1MmJkMDZkNjIyMGNhMWIiLCJwIjoiaiJ9) to figure out buttons. The button that adds a new element to the collection will need to know what its action is. Similarly, the button that removes a specific element from the collection will need to know what its action is.

As a first pass, the behavior of buttons can be hardcoded based on where they're used (add inserts an element, remove removes a specific element, cancel closes a specific modal, etc.). 

> ```purescript
>         , template: -- `contents` replaced by `template`
> ```

I don't think we need to replace `contents`, but adding `template` can be one way to address the button concern. As mentioned above, if we have the template, we can hardcode the behavior of what the add button does.

> Obviously it's not perfect, putting an array of `Field` inside `array_` being an obvious problem, but I hope the idea is a bit clearer.

If the purpose of these collections is to house what are effectively multiple sections, why not make the `template` a single section and we can keep the `contents` as an array of sections? We shouldn't have to add any new values or inputs or whatever if that's all that we're doing.

citizengabe: As currently implemented, this would require that `contents` be changed from `NonEmpty Array f` to `Array (NonEmpty Array f)` for `Collection` to accommodate the nested arrays.

If we need to change the type anyway, can't we do `NonEmpty Array (NonEmpty Array f)` with a default section and avoid `template`?

```haskell
{ name: "Users Form"
, contents:
  [ { name: "Users"
    , link: "users"
    , contents:
      [ Collection
        { name: "User"
        , key: "users"
        , contents
          [ [ { name: "Name"
            , key: "name"
            , input: Text
              { value: NotSet
              ...
              }
            }
          , { name: "Email"
            , key: "email"
            , input: Text
              { value: NotSet
              ...
              }
            }
          ] ]
        }
      ]
    }
  ]
}
```

joneshf-cn: > As currently implemented, this would either require that `contents` be changed from `NonEmpty Array f` to `Array (NonEmpty Array f)` for `Collection` to accommodate the nested arrays.

What I'm suggesting is that we go from:
```PureScript
type Page a
  = { name :: String
    , contents :: NonEmpty Array (Tab a)
    }

type Tab a
  = { name :: String
    , link :: String
    , contents :: NonEmpty Array (Section a)
    }

type Section a
  = { name :: String
    , contents ::
        NonEmpty
          Array
          { name :: a
          , visibility :: a
          , description :: a
          , key :: Key
          , input :: Input a
          }
    }
```
to
```PureScript

type Page' a
  = { name :: String
    , contents :: NonEmpty Array (Tab' a)
    }

type Tab' a
  = { name :: String
    , link :: String
    , contents ::
        NonEmpty Array
          ( Either
            (Section a)
            { name :: String
            , contents :: Array (Section a)
            , template :: Section a
            }
          )
    }

type Section a
  = { name :: String
    , contents ::
        NonEmpty
          Array
          { name :: a
          , visibility :: a
          , description :: a
          , key :: Key
          , input :: Input a
          }
    }
```

Or to:
```PureScript
type Page' a
  = { name :: String
    , contents :: NonEmpty Array (Tab' a)
    }

type Tab' a
  = { name :: String
    , link :: String
    , contents :: NonEmpty Array (CollectionOrSection a)
    }

data CollectionOrSection a
  = Collection
    { name :: String
    , contents :: Array (Section a)
    , template :: Section a
    }
  | Section (Section a)

type Section a
  = { name :: String
    , contents ::
        NonEmpty
          Array
          { name :: a
          , visibility :: a
          , description :: a
          , key :: Key
          , input :: Input a
          }
    }
```
Or whatever else is equivalent.

Zooming in, I'm suggesting this change:
```diff
-    , contents :: NonEmpty Array (Section a)
+    , contents ::
+        NonEmpty Array
+          ( Either
+            (Section a)
+            { name :: String
+            , contents :: Array (Section a)
+            , template :: Section a
+            }
+          )
```
Or its equivalent.

> If we need to change the type anyway, can't we do `NonEmpty Array (NonEmpty Array f)` with a default section and avoid `template`?

`NonEmpty f a` is always equivalent to `{ head :: a, tail :: f a }`, so we can always go between the two. Using [ad-hoc types][] should make the intent of the data structure clearer because we can provide a domain-specific names to the idea behind what the `NonEmpty Array _` represents. Using domain specific names tends to lead to a feedback loop where [edge cases are surfaced][edge cases] and the overall code becomes [more correct to the problem][correct].

[ad-hoc types]: https://jaspervdj.be/posts/2016-05-11-ad-hoc-datatypes.html
[correct]: https://github.com/citizennet/purescript-lynx/pull/6
[edge cases]: https://citizennet.slack.com/archives/C3Q0ATZST/p1551470391098500

davezuch: A couple questions arise if we go with `contents :: Array (Section a)`. 

We lose the ability to represent `NotSet` for arrays. Is that okay? That prevents us from providing default values for arrays. A form creator can recover most of that functionality just by setting default values on the fields within the array, which arguably is preferable, but we still lose the ability to say the default value for some array is two values, one of _foo_, the second of _bar_. Something we currently do for targeting sources on campaigns, where [the first is _Core Interests_, the second is _Affinity Interests_](https://staging.citizennet.com/manager/#/groups/facebook/162454/campaigns/new/targeting). Maybe arrays have an additional configuration of starting items, and we can use that plus `Input` defaults to fully recover that functionality?

Additionally, if we go with the `template` solution for defining our schema outside of `contents` (which I think we have to do for this approach), we have the problem of allowing invalid state to be represented, as our `Input`s within our template have a `value`, which, doesn't really make sense as they can't be anything but `NotSet`, in which case, they probably just shouldn't be in there.

At least for `template` I think we would want to decouple our values from our layout/schema definition.

joneshf-cn: > We lose the ability to represent `NotSet` for arrays. Is that okay?

I'm not sure why we need the ability to express that.

The only time `NotSet` is used is for validation. It doesn't seem to make sense to attempt to validate an array as an entity separate from its elements. Validating the individual elements of an array makes sense. But it seems like wanting to have `NotSet` on a collection of things is tantamount to wanting to have `NotSet` on a `Page _` (which we don't support).

> That prevents us from providing default values for arrays. A form creator can recover most of that functionality just by setting default values on the fields within the array, which arguably is preferable, but we still lose the ability to say the default value for some array is two values, one of _foo_, the second of _bar_. Something we currently do for targeting sources on campaigns, where [the first is _Core Interests_, the second is _Affinity Interests_](https://staging.citizennet.com/manager/#/groups/facebook/162454/campaigns/new/targeting). Maybe arrays have an additional configuration of starting items, and we can use that plus `Input` defaults to fully recover that functionality?

All of this feels like something we can do when we're ready to support that. For MVP, we don't need any of this functionality. Can we not move forward without addressing this?

> That prevents us from providing default values for arrays. 

> Additionally, if we go with the `template` solution for defining our schema outside of `contents` (which I think we have to do for this approach), we have the problem of allowing invalid state to be represented, as our `Input`s within our template have a `value`, which, doesn't really make sense as they can't be anything but `NotSet`, in which case, they probably just shouldn't be in there.

Sorry, I'm not understanding this point. Why must the `InputSource _` be `NotSet` for all of the `value`s in a `template`? Why can't the `template` specify default values for each element that is added?

davezuch: > The only time `NotSet` is used is for validation.

We also use it for determining whether to resort to a default value.

> It doesn't seem to make sense to attempt to validate an array as an entity separate from its elements.

We'll definitely want the ability to say there must be a minimum of `n` array items. That's separate from validating the values of that array item. In my User example above, `name` and `email` might both be required, but that only means that we don't have a valid User without them, which is separate from requiring we have 1 or more Users.

> All of this feels like something we can do when we're ready to support that. For MVP, we don't need any of this functionality. Can we not move forward without addressing this?

That's an option, but I think it's useful to think about these things now so that we don't get ourselves too deep in a flawed design. That was one of the things that bit us in our last form library, we were having to bandaid on solutions for use cases it was never designed to support. Granted, refactoring is a lot cheaper in PS, but if we can save ourselves some pain down the road I'd like to.

> Sorry, I'm not understanding this point. Why must the `InputSource _` be `NotSet` for all of the `value`s in a `template`? Why can't the `template` specify default values for each element that is added?

I don't think that makes sense semantically. `InputSource a` is a sum of `UserInput a | Invalid a | UserCleared | NotSet`, it's supposed to represent values that are a result of user interaction with the form. We use `default :: Maybe Expr` to construct a default value in the absence of valid user input.

In the same vein, doesn't really make sense to have `errors` in our template. Doesn't make sense to have them in `Page Expr` either, though we've discussed that at length in a [previous PR](https://github.com/citizennet/purescript-lynx/pull/16#discussion_r268286652) (which, I would like to circle back to at some point).

joneshf-cn: > That's an option, but I think it's useful to think about these things now so that we don't get ourselves too deep in a flawed design. That was one of the things that bit us in our last form library, we were having to bandaid on solutions for use cases it was never designed to support. Granted, refactoring is a lot cheaper in PS, but if we can save ourselves some pain down the road I'd like to.

This is something that's been in the back of my mind for a while about accountability and autonomy. I mentioned that I wanted to [write some stuff about it][write up]. But this PR isn't the right avenue for it. I'll put something in the slack.

[write up]: https://citizennet.slack.com/archives/CGPP2STC4/p1560203811122000

> I don't think that makes sense semantically. `InputSource a` is a sum of `UserInput a | Invalid a | UserCleared | NotSet`, it's supposed to represent values that are a result of user interaction with the form. We use `default :: Maybe Expr` to construct a default value in the absence of valid user input.

Oh right. Sorry about that.

citizengabe:
Closing in favor of #37

PR #36, opened 2019-06-05
Start using nix

joneshf-cn:
## What it does

While helping Gabe get setup, we were running into version issues with different things. When we got to this repo, we ran into a version issue with `make`. Since we'd been hitting version differences throughout other projects, we tried to fix this going forward for this repo. We use `nix` specify the dependencies of all the programs we need in this repo. Ideally this should mean that anyone that comes to the project will always have the correct versions of everything.

The motivation behind `nix` is that there are some programs we need available in development. This doesn't negate the need/use of `docker` or `k8s`.

## Review asks

* Are we okay with adding `nix` to the repo? Dave and I have talked about it in the past and we're both onboard. Gabe is also onboard. Any thing to be concerned about?

## TODO

* [ ] Document `nix`
    * [ ] Installation
    * [ ] Motivation
* [ ] Document `direnv`
    * [ ] Installation
    * [ ] Motivation

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,22 @@
+set -euo pipefail
+
+use_nix() {
+    local cache=.direnv.cache
+    if [[ ! -e "${cache}" || \
+              "${HOME}/.direnvrc" -nt "${cache}" || \
+              .envrc -nt "${cache}" || \
+              default.nix -nt "${cache}" \
+        ]]; then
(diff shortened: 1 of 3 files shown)


joneshf-cn:
I'm not going to get to this anytime soon. Will make a new branch if I want to push this again.

PR #37, opened 2019-06-12
Add section sequence

citizengabe:
## What it does

Please see https://github.com/citizennet/purescript-lynx/pull/35#discussion_r291409630 for background.

This PR implements https://citizennet.atlassian.net/browse/PS-3013 and supersedes #35 (it was easier to start over from `master` than to undo the previous work).

In order to more easily _visually_ distinguish between the various types in the `Lynx.Form` hierarchy, I've changed the name of the `contents` property to more closely match the type of the children.

I've also changed the child of a `Tab a` to be `Either (Section a) (Sequence a)` allowing for multiple duplicate sections.

## Review asks

@davezuch @joneshf-cn I'd like to discuss the `template` property which I currently have as `Unit`. I want to get this PR now though but will try to distill the conversation from https://github.com/citizennet/purescript-lynx/pull/35#discussion_r291409630 this afternoon.


src/Lynx/Form.purs
```diff
@@ -30,17 +30,19 @@ import Test.QuickCheck (class Arbitrary)
 import Test.QuickCheck.Arbitrary (genericArbitrary)
 import Type.Row (type (+))
 
-type LayoutRows c r =
```
joneshf-cn, line 33: 😻 Inlining this seems much clearer what the different pieces are.

src/Lynx/Form.purs
```diff
@@ -30,17 +30,19 @@ import Test.QuickCheck (class Arbitrary)
...
-  ( name :: String
-  , contents :: NonEmpty Array c
-  | r
-  )
 
-type Page f = Record (LayoutRows (Tab f) ())
+type Page f = { name :: String
+              , tabs :: NonEmpty Array (Tab f)
```
joneshf-cn: We should recognize that changing any of these fields is a breaking change.

Since we rely on the instances for decoding/encoding that depend on the labels of a record and because we store `Page _`s directly as those JSON blobs in OMS, this means that we'll not be able to parse older versions of `Page _` going forward. E.g. all of the data in all current deployments of OMS will be no longer be parseable.

Nothing is live, so we can probably just break stuff. But, we might want to think about no longer storing data as JSON blobs in OMS sooner rather than later. Either that or thinking about migrating through different versions.

test/Test/Lynx/Form.purs
```diff
@@ -192,8 +198,10 @@ testTab :: String
 testTab = """
   { "name": "User"
   , "link": "user"
-  , "sections":
-    [ """ <> testSection <> """
+  , "contents":
+    [ { "tag": "Left"
```
joneshf-cn: For consistency with other sum types we have, maybe we should make a data type for `TabContents _` and follow the pattern of stuffing a `type` field into the JSON object to distinguish the cases. How do you feel about that?

src/Lynx/Form.purs
```diff
@@ -30,19 +33,30 @@ import Test.QuickCheck (class Arbitrary)
...
+  { name :: String
+  , fields :: NonEmpty Array (Field f)
+  }
 
-type Section f = { name :: String
-                 , fields :: NonEmpty Array (Field f)
-                 }
+type Sequence f =
+  { name :: String
+    -- TODO: need to figure out what `template` should be
```
joneshf-cn: Based on the discussion in #35, it seems like it would be somehting like:
```PureScript
type Sequence a
  = { name :: String
    , template :: NonEmpty Array (Template a)
    , sections :: Array (Section a)
    }

type Template a
  = { name :: a
    , visibility :: a
    , description :: a
    , key :: Key
    , input :: TemplateInput a
    }

data TemplateInput a
  = TemplateCurrency
    { default :: Maybe a
    , required :: a
    , placeholder :: a
    }
  | TemplateDateTime
    { default :: Maybe a
    , required :: a
    , placeholder :: a
    }
  | TemplateDropdown
    { default :: Maybe a
    , required :: a
    , options :: a
    , placeholder :: a
    }
  | TemplateText
    { default :: Maybe a
    , required :: a
    , placeholder :: a
    , maxLength :: Maybe a
    , minLength :: Maybe a
    }
  | TemplateToggle
    { default :: Maybe a
    }
  | TemplateTypeaheadSingle
    { default :: Maybe a
    , options :: a
    , resultValue :: a
    , results :: a
    , uri :: a
    }
```

That seem right?

joneshf-cn (approved):
Looks good to me.

I'd be happy merging this whenever you're comfortable with the `template` stuff. Could even merge this and address `template` in a follow-up PR.

src/Lynx/Form.purs
citizengabe: I agree that we can probably just blow away what we have in OMS right now, but think we need a longer-term migration strategy as well.

test/Test/Lynx/Form.purs
citizengabe: That was my intention in the next iteration (you mentioned the utility of controlling the encoding while we were working on it 😄). I plan to write something similar to what I did in #35.

citizengabe:
@joneshf-cn I'd prefer to at least do a first pass on the template before merging. But I plan to get that done today.

src/Lynx.purs
```diff
@@ -140,20 +140,29 @@ component =
...
     where
-    byLink :: Tab ExprType -> Boolean
-    byLink { link } = activeTab == link
+      byLink :: Tab ExprType -> Boolean
+      byLink { link } = activeTab == link
 
-    tab :: Tab ExprType
-    tab = fromMaybe (Data.NonEmpty.head tabs) (find byLink tabs)
+      tab :: Tab ExprType
+      tab = fromMaybe (Data.NonEmpty.head tabs) (find byLink tabs)
```
davezuch: We've been following the standard of not indenting `where` clauses.

src/Lynx.purs
```diff
@@ -140,20 +140,29 @@ component =
...
       [ Format.subHeading_ [ HH.text section.name ] ]
-      $ Data.Array.fromFoldable $ renderField <$> section.contents
+      $ Data.Array.fromFoldable $ renderField <$> section.fields
+
+  renderSequence :: Sequence ExprType -> H.ParentHTML Query (ChildQuery m) ChildSlot m
+  renderSequence sequence =
+    Card.card_ $
+    append
+    [ Format.subHeading_ [ HH.text sequence.name ] ]
+    $ Data.Array.fromFoldable $ do
```
davezuch: I don't think we need `fromFoldable` here. It's used elsewhere to go from `NonEmpty Array _` to `Array _`, but `sequence.sections` is already `Array _`.

src/Lynx/Form.purs
```diff
@@ -30,17 +33,30 @@ import Test.QuickCheck (class Arbitrary)
...
+type Section f =
+  { name :: String
+  , fields :: NonEmpty Array (Field f)
+  }
+
+type Sequence f =
+  { name :: String
+    -- TODO: need to figure out what `template` should be
+  , template :: Unit
+  , sections :: Array (Section f)
```
davezuch: I think the naming of these labels is subtly misrepresentative. I don't want to be pedantic, but I'd like to avoid unclear naming, as that can lead to confusion and misunderstanding. In our other layout structures, we use this label (formerly `contents`) to denote a collection of distinct layout elements that we want at this level. Here, `template` is more representative of that distinct layout element within its parent's `contents`, while `sections` is a bit different from any label we currently have, but I think a name like `values` might be clearer.

So what I'm suggesting is:
```diff
-  , sections :: Array (Section f)
+  , values :: Array (Section f)
```

In fact, in the spirit of other changes in this PR, I think it'd be clearer to also update `Tab`'s label from `contents` to `sections`, because I see `Sequence` here as just being a special `Section`. Open to discussing though, if anyone sees things differently.

davezuch (commented):
This is looking good so far! I just have a few comments below

src/Lynx/Form.purs
davezuch: That looks great @joneshf-cn! I would just suggest a few adjustments to the first two types: I think `key`, `visibility`, `description` belong on `Sequence` instead of `Template` (`visibility` and `description` should also be on other layout types, but that's one of those things that have been punted, on as they don't impact MVP). Not sure we even need `Template` actually. I think we could make it `template :: Section a TemplateInput` if we assume that the extra type param somehow tells the child fields that inputs should be `TemplateInput a` instead of `Input a`.

Combined with the suggestion from my code review, we'd end up with something more like:

```purescript
type Sequence a
  = { name :: String
    , template :: Template (Section a)
    , key :: Key
    , values :: Array (Section a)
    -- we could even leave these last two out
    , visibility :: a
    , description :: a
    }
```

Maybe we'll want to work on our `Input` rows a bit, so we can have something like:

```purescript
type BaseInputRows f r =
  ( default :: Maybe f
  , errors :: Errors ValidationError
  | r
  )

data BaseInput f r
  = Currency (Record (BaseInputRows f + CurrencyRows f + r))
  | DateTime (Record (BaseInputRows f + DateTimeRows f + r))
  ...

type TemplateInput f = BaseInput f ()
type Input f = BaseInput f (value :: InputSource f)
```

src/Lynx/Form.purs
citizengabe: > Here, template is more representative of that distinct layout element within its parent's contents, while sections is a bit different from any label we currently have, but I think a name like values might be clearer.

I disagree. I think `sections` still represents layout elements as `contents` did before; they're just not _necessarily_ explicitly laid out. They could be though. `template` OTOH is just that, a _template_ for a layout element.

I'm fine with changing `Tab`'s `contents` to `sections` and sharing that property w/ `Sequence` though. The JSON representation of `TabContents` should be sufficient for orientation.

src/Lynx/Form.purs
citizengabe: Should `Sequence` have `minLength` and `maxLength`?

src/Lynx/Form.purs
davezuch: > I think `sections` still represents layout elements as `contents` did before; they're just not _necessarily_ explicitly laid out. They could be though.

Unless we're talking about different things, they can't be though. The value for `sections` here is not something we define on the form definition, it is the result of user input/interaction with the form. In fact, now that I think about it, the type should probably be `values :: UserInput (Array (Section f))` or even `values :: UserInput (Array (Section ExprType))`.

My point is `contents` represented _discrete_ layout elements, while `sections` here (what I'm proposing be named `values`) represents our array _values_ for _one_ of those layout elements.

That is, before we had:
```purescript
tabUser = Tab
  { contents: -- schema + values
    [ sectionName
    , sectionInterests
    , sectionBackground
    ]
  ...
  }

sectionName = Section
  { contents: -- schema + values
    [ fieldFirst
    , fieldMiddle
    , fieldLast
    ]
  ...
  }
```

Now we have:
```purescript
tabUser = Tab
  { contents: -- schema + values
    [ sectionOrSequenceName
    , sectionOrSequenceInterests
    , sectionOrSequenceBackground
    ]
  ...
  }

sectionOrSequenceName = Sequence
  { template: templateSectionName -- schema
  , values: UserInput -- values (could also be `NotSet`)
    [ valueSectionName
    , valueSectionName
    , valueSectionName
    , valueSectionName
    ]
  ...
  }

templateSectionName = TemplateSection
  { fields: -- schema
    [ fieldFirst
    , fieldMiddle
    , fieldLast
    ]
  ...
  }
```
In both cases we're wrapping a `Section` in an array, but those arrays represent different things, they behave and function differently, so their naming should also be distinct.

src/Lynx/Form.purs
davezuch: > Should `Sequence` have `minLength` and `maxLength`?

Eventually, we don't need to support it right now.

davezuch: Actually, I definitely think we should use the `BaseInputRows` idea. We want to make sure that `Input` and `TemplateInput` stay consistent with each other, this would lower the maintenance burden.

citizengabe: I can factor the commonalities out once we have everything working. I'd like to do it in a separate PR though.

src/Lynx/Form.purs
citizengabe: I think some of my confusion is over the type of `Sequence.values`. You have it as `InputSource (Array (Section a))`, while I have it as `Array (Section a)`.

src/Lynx/Form.purs
davezuch: That's fine, I could do it later as well. I'd also like to clean up those rows, I think we can shrink down the different rows we're combining to at most 2 per input type, but yeah, factoring this certainly isn't a showstopper.

src/Lynx/Form.purs
citizengabe: I suppose since we're adding elements based on user input, your way makes sense.

citizengabe: > The value for sections here is not something we define on the form definition, it is the result of user input/interaction with the form.

I had in mind a situation where we had a default section in the sequence (I thought you had mentioned a use case for that).

citizengabe: Maybe we need a `default` property for that. I don't know if that use case is required for the MVP.

davezuch: Yeah, this is why I'm starting to wonder whether the child form approach may have been clearer or more representative of what we're trying to achieve. I think it's worthwhile to think of `Sequence` as another `Input` type, like `Toggle` or `Text`, it just lives at a different level in our layout structure, but it still has a `key`, and a `value`. The `InputSource _` type is the type we wrap possible user values in, which is what the values of our `Sequence` are.

davezuch: > Maybe we need a `default` property for that. I don't know if that use case is required for the MVP.

Yeah, I would say we can leave that out for now.

joneshf-cn (approved):
Took a look at everything so far. Still looking good!

src/Lynx/Form.purs
```diff
@@ -30,17 +30,135 @@ import Test.QuickCheck (class Arbitrary)
...
 
-type Page f = Record (LayoutRows (Tab f) ())
+type Tab f =
+  { name :: String
+  , link :: String
+  , sections :: NonEmpty Array (TabSections f)
+  }
 
-type Tab f = Record (LayoutRows (Section f) (link :: String))
+data TabSections f = TabSection (Section f)
```
davezuch: Can we bump this down to the next line? Prefer not to have ad-hoc indentation.

src/Lynx/Form.purs
```diff
@@ -30,17 +30,135 @@ import Test.QuickCheck (class Arbitrary)
...
+
+instance decodeJsonTabSections :: DecodeJson (TabSections Expr) where
+  decodeJson json = do
+    x <- decodeJson json
+    x .: "type" >>= case _ of
+      "TabSection" -> pure <<< TabSection <=< decodeJson $ json
+      "TabSequence" -> pure <<< TabSequence <=< decodeJson $ json
+      t -> Left $ "Unsupported TabSections type: " <> t
+
+type Template f =
```
davezuch: Can we rename this to `TemplateSection` for clarity?

src/Lynx/Form.purs
```diff
@@ -269,18 +388,93 @@ instance arbitraryValidationError :: Arbitrary ValidationError where
...
-    contents <- traverse evalSection tab.contents
-    pure tab { contents = contents }
+  evalTab tab =
+    tab { sections = _ } <$> traverse evalTabSections tab.sections
+
+  evalTabSections :: TabSections Expr -> Either EvalError (TabSections ExprType)
+  evalTabSections =
+    tabSections
+      ((map TabSection) <$> evalSection)
+      ((map TabSequence) <$> evalSequence)
```
davezuch: Are the extra parens necessary? Don't feel like they should be.

src/Lynx/Form.purs
```diff
@@ -527,40 +736,58 @@ asyncFromTypeahead typeahead x = case toString typeahead.uri of
...
+                                   , maxLength: Nothing
+                                   , minLength: Nothing
+                                   }
+                }
+    , values:
+        UserInput [ { name: "Social Creative"
+                    , fields: Data.NonEmpty.singleton mvpSocialAccount
+                    }
+                  ]
+    }
```
davezuch, line 793: Can we fix this indentation so we're not using ad-doc spacing? See surrounding vals, e.g. `mvpDetailsSection`

davezuch (approved):
Just a few minor requests in the comments, but looks good!

PR #38, opened 2019-06-17
PS-3013 Carry more info. in `TemplateSection _`

joneshf-cn:
## What it does

We attempted to use the templates to stamp out a value, but ran into an issue. We need a bit more information in `TemplateSection _`, we only had enough information to create `Input _`s. We're creating entire `Section _`s when we stamp out templates, so we need enough information to fill out the entirety of a `Section _`. That means we need an equivalent amount of information as a `Section _`.

`Section _`s have a `name :: String` and a `fields :: NonEmpty Array (Field _)`. We had `name :: String` and a `fields :: NonEmpty Array (TemplateInput _)`, which was equivalent to have a template for `Field _`, not a template for `Section _`.

We extend the templates to be equivalent to a `Section _`.

## Review asks

* This PR doesn't attempt to connect all of the dots for PS-3013. It makes a small incremental change based on what we speculate should be the next step. We attempted to handle all of the edges in #35 and #37, yet we were still blind to this case. If the overall idea here makes sense and the code isn't blatantly broken, can we merge this through without attempting to force out all of the problems with it and continue this iterative process of finishing PS-3013?

### DIFF
--- a/dist/main.js
+++ b/dist/main.js
@@ -20550,8 +20550,8 @@ var PS = {};
       };
   };
   var singletonError = function (dictOrd) {
-      return function ($580) {
-          return Errors(Data_Set.singleton($580));
+      return function ($588) {
+          return Errors(Data_Set.singleton($588));
       };
   };
   var semigroupErrors = function (dictOrd) {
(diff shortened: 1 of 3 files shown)


joneshf-cn:
Requesting a review from Dave, Gabe, and Vance, though I think one should be sufficient.

joneshf-cn:
I was going to request a review, but something is currently broken in GitHub. Mentioned on Slack.

joneshf-cn:
Uhh, apparently I forgot how to GitHub. You can mention people in an issue 😅.

cc @davezuch, @citizengabe, @vanceism7.

citizengabe (approved):
This all looks correct and follows from our discussion on Friday.

joneshf-cn:
Thanks!

PR #39, opened 2019-06-17
PS-3013 Add a way to stamp out templates

joneshf-cn:
## What it does

We add a function that should create templates in sequence when asked for them.

Paired with Gabe through this.

## Review asks

* We still need to hook up the render. That ws split out to PS-3014. It seems like we might have everything we need here. Is there anything blantly missing?

### DIFF
--- a/dist/main.js
+++ b/dist/main.js
@@ -20545,13 +20545,13 @@ var PS = {};
               if (v instanceof TabSequence) {
                   return g(v.value0);
               };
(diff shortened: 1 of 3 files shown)


davezuch (approved):
These updates look good to me

citizengabe (approved):
LGTM

joneshf-cn:
Thanks for the reviews!

PR #40, opened 2019-06-17
Add `make format`

citizengabe:
## What it does

In an effort to make our lives easier, Hardy and I have added a new task to `Makefile`: `format`.

`format` runs [`purty`](https://gitlab.com/joneshf/purty) on `.purs` files in `src/`, `test/`, and `examples/`.

Because `purty` 4.x uses PureScript's new CST parser (which breaks backward compatibility) `if` expressions were broken as formatted. We took this opportunity to change these to pattern guards or case statements where we could. We also used language-level syntax in place of userland operators in a few places to reduce indentation.

## Review asks

I think things look good and I'd love to stop thinking about formatting, but some of the changes weren't strictly necessary to integrate `purty`. In particular, I'm requesting feedback on the syntax changes more than the formatting.


vanceism7 (approved):
The formatting is in someways different from what I'm used to, but I think in general, the syntax/formatting changes do ultimately aid in better readability, so I like it :+1:

Makefile
```diff
@@ -139,7 +139,7 @@ dist/main.js: $(BUILD)/main.js
 	npx browserify $< --outfile $@
 
 .PHONY: format
-format: $(FORMAT_EXAMPLES) $(FORMAT_TESTS) $(FORMAT_TESTS)
+format: $(FORMAT_EXAMPLES) $(FORMAT_TESTS) $(FORMAT_SRCS)
```
joneshf-cn, line 142: Sorry about that.

Makefile
```diff
@@ -105,11 +111,16 @@ $(BUILD)/test.out: $(BUILD)/test.js | $(BUILD)
...
+	npx purty --write $<
+	@mkdir -p $$(dirname $@)
+	@touch $@
+
 $(NODE_MODULES): package.json
 	npm install
 	touch $@
 
-$(OUTPUT)/Main/index.js: $(EXAMPLES) $(SRCS) $(BOWER_COMPONENTS) $(NODE_MODULES) | $(BUILD)
+$(OUTPUT)/Main/index.js: format $(EXAMPLES) $(SRCS) $(BOWER_COMPONENTS) $(NODE_MODULES) | $(BUILD)
```
joneshf-cn:
```suggestion
$(OUTPUT)/Main/index.js: $(FORMAT_EXAMPLES) $(FORMAT_SRCS) $(BOWER_COMPONENTS) $(NODE_MODULES) | $(BUILD)
```

Looking at the whole `Makefile`, we're probably better off if we talk about individual files that should exist rather than a phony target. It's more granular, and should be more accurate.

Makefile
```diff
@@ -105,11 +111,16 @@ $(BUILD)/test.out: $(BUILD)/test.js | $(BUILD)
...
 	npx psa $(PSA_ARGS) $(RTS_ARGS) $(DEPS) $(EXAMPLES) $(SRCS)
 
 $(OUTPUT)/Test.Main/index.js: $(EXAMPLES) $(SRCS) $(TESTS) $(BOWER_COMPONENTS) $(NODE_MODULES) | $(BUILD)
```
joneshf-cn:
```suggestion
$(OUTPUT)/Test.Main/index.js: $(FORMAT_EXAMPLES) $(FORMAT_SRCS) $(FORMAT_TESTS) $(BOWER_COMPONENTS) $(NODE_MODULES) | $(BUILD)
```

Similar to above, if we talk about which things we really need to depend on, it should keep things more accurate.

joneshf-cn: Side note, I'm not entirely sure why it's trying to remove a blank space at the start of the line. Hopefully this doesn't break the file.

joneshf-cn (approved):
You know I approve of this!

joneshf-cn:
Another thing to note is that given how we've setup our `Makefile`, if you run `make watch`–or really, almost any of the targets–it will run the formatter after each save but before compiling. It's not quite the same as having direct editor integration that runs before saving. But from 10,000 feet, it's about the same.

example/Lynx/Page/Form.purs
```diff
@@ -108,11 +107,11 @@ component =
...
-eval
-  :: ∀ m
-   . MonadAff m
-  => Query
-  ~> H.ParentDSL State Query ChildQuery ChildSlot Message m
+eval ::
+  ∀ m.
+  MonadAff m =>
+  Query
+    ~> H.ParentDSL State Query ChildQuery ChildSlot Message m
```
davezuch, line 114: Not gonna lie, it hurts a little to move these symbols to the end, though I'll live. @joneshf-cn did make a good case for them in a [previous PR comment](https://github.com/citizennet/purescript-lynx/pull/20#discussion_r270604925).

Is there nothing that can be done about this `~>` though? Shouldn't this be:
```purescript
eval ::
  ∀ m.
  MonadAff m =>
  Query ~>
  H.ParentDSL State Query ChildQuery ChildSlot Message m
```

davezuch: Although, I must add, everywhere else we put our punctuation at the start of a newline, instead of ending our line with it.
```purescript
-- no
val = [
  "foo",
  "bar",
  "baz"
]

-- yes
val =
  [ "foo"
  , "bar"
  , "baz"
  ]

-- no 
val2 =
  wrap $
    first <>
    second <>
    third

-- yes 
val2 =
  wrap
    $ first
    <> second
    <> third

-- no
type MyType =
  FooBar \/
  Biz \/
  Bazel

-- yes
type MyType
  = FooBar
    \/ Biz
    \/ Bazel

-- yes?
myFunc ::
  forall a.
  Show a =>
  a ->
  String

-- no?
myFunc
  :: forall a
   . Show a
  => a
  -> String
```

test/Test/Lynx/Form.purs
```diff
@@ -201,51 +209,74 @@ testPageEither :: Either String (Page Expr)
...
-    , """ <> active <> """
+    [ """
+    <> firstName
+    <> """
+    , """
+    <> lastName
+    <> """
+    , """
+    <> active
+    <> """
```
davezuch, line 576: I guess it's not a big deal, but is this something we could configure? All these JSON strings are a bit less readable now

davezuch (approved):

example/Lynx/Page/Form.purs
joneshf-cn: > Is there nothing that can be done about this `~>` though? Shouldn't this be:
> 
> ```purescript
> eval ::
>   ∀ m.
>   MonadAff m =>
>   Query ~>
>   H.ParentDSL State Query ChildQuery ChildSlot Message m
> ```

`(~>)` is a [user-land operator][nattrans]. All user-land operators are formatted with the same rules: if the entire thing spans multiple lines, format the left operand, indent the operator, then format the right operand; otherwise, format the whole thing over a single line. We (CitizenNet) can put it on a single line if that's more appealing.

I'm not against a different formatting for user-land operators if it leads to less noisy diffs and doesn't special case anything in the user-land. But, it'll be applied uniformly to all user-land operators. Feel free to [open an issue][new issue]

[nattrans]: https://pursuit.purescript.org/packages/purescript-prelude/4.1.1/docs/Data.NaturalTransformation#t:type%20(~>)
[new issue]: https://gitlab.com/joneshf/purty/issues/new

test/Test/Lynx/Form.purs
joneshf-cn: What did you have in mind?

PR #42, opened 2019-06-18
PS-3013 Unstamp `Section _`s

joneshf-cn:
## What it does

Adds a way to remove a `Section _` that's been stamped out in a `Sequence _`.

## Review asks

* As with #39, this PR only implements the logic. The UI is being held off until a further PR. Is that okay? Is there anything blatant that we're missing?


davezuch (approved):
Looks good!

PR #43, opened 2019-06-18
PS-3014 Render `Remove` buttons for `Sequence _` `Section _`s

joneshf-cn:
## What it does

To use the functionality of removing a `Section _` from a sequence, we add a button to remove any element individually.

<details>
<summary>`Remove` buttons</summary>

Screen Shot 2019-06-18 at 4 23 13 PM
Screen Shot 2019-06-18 at 4 23 19 PM
</details>

## Review asks

* Any blatant problems that were missed?
* This looks to be the last bit of PS-3014, is there anything else left?


src/Lynx.purs
```diff
@@ -174,6 +182,24 @@ component =
...
+          [ HH.text section.name
+          , Button.buttonClear
+            [ css "absolute pin-r pr-0"
+            , HE.onClick (HE.input_ $ RemoveSection key index)
+            ]
+            [ Icon.close_
+            ]
+          ]
+        ]
+        <> Data.Array.fromFoldable (map renderField section.fields)
```
citizengabe, line 199: That index needs to be propagated down to `renderField`. I'm working in the same area in #45.

Maybe we should work together on this one or at least coordinate the order of merges.

citizengabe: Per our conversation in Slack, I think we should merge this and I'll pull in the changes to #45 and work from there.

citizengabe (approved):

PR #45, opened 2019-06-18
PS-3013 Fix stamped field ids

citizengabe:
## What it does

In order to avoid `id` collisions for DOM elements in sections stamped out from templates, I'm building the keys using the indices of `TemplateSequence.values`.


citizengabe:
I'm leaving this in draft status until I get some UI testing in place.

src/Lynx.purs
```diff
@@ -140,29 +140,29 @@ component =
...
+      $ Data.Array.fromFoldable $ renderField (map show index) <$> section.fields
 
   renderSequence :: Sequence ExprType -> H.ParentHTML Query (ChildQuery m) ChildSlot m
   renderSequence sequence =
     Layout.container_
       ( [ Format.subHeading_ [ HH.text sequence.name ] ]
         <> case sequence.values of
-            UserInput sections -> renderSection <$> sections
-            Invalid sections -> renderSection <$> sections
+            UserInput sections -> Data.Array.mapWithIndex (renderSection <<< Just) sections
```
citizengabe: Should I be using `FunctorWithIndex` here?

joneshf-cn: I don't think it much matters. They ought to be equivalent if not the same.

citizengabe: I ended up changing it anyway

joneshf-cn (approved):
Looks good!

PR #46, opened 2019-06-19
Minify dist/main.js to hopefully avoid some merge conflicts

citizengabe:
## What it does

Stops tracking `dist/main.js` (a generated file) in order to prevent merge conflicts and noise in PRs.

I've found on a few occasions that I'll get merge conflicts when rebasing onto master locally. I've also needed to merge a PR manually simply to update this file.


joneshf-cn (approved):
I'm down with this.

joneshf-cn (changes requested):
Actually, I think this is going to break our CI caching: https://circleci.com/gh/citizennet/purescript-lynx/196?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link. The overall build is still passing, but the cache restoration step is failing because `dist/main.js` doesn't exist anymore. We rely on `dist/main.js` for part of the cache key. We'll have to change that as part of this PR as well.

The change could be to alter what we cache and how, it could be to use a different file for caching, or it could be something else. Whatever you think is an appropriate change.

citizengabe:
What we're caching looks legit, but maybe we could generate a key associated w/ a particular run of `make` and put that in a file. Then we could `checksum` that in `.circleci/config.yaml`.

joneshf-cn:
Give it a whirl!

joneshf-cn:
Hmm, is that equivalent to minifying `dist/main.js`? The cache key would need to be kept up to date with every commit. Any diffs would be one-line (does that mean no merge conflicts?). That seems like nearly the same as what would happen if we minified `dist/main.js`. It would be one-line so we'd end up with a similar story for diffs (and merge conflicts, by extension).

citizengabe:
😆 that's the same conclusion I came to last night. Any merge conflict on that one line would indicate the need for a rebuild and invalidating the cache.

I'll try running it through `uglifyjs` and we can see how that works for us.

joneshf-cn:
I also realized that we're not really caching anything. Our cache keys are jacked. Maybe we'll get more benefit from simplifying what we cache? We could do one cache for both `bower_components` and `node_modules`? I think a clean build (with dependencies installed) only takes a minute or two. Most often, CI completes in the amount of time it takes to switch from terminal/editor to GitHub and open a PR. It's another idea, not necessarily something we have to do.

citizengabe:
I was about to push up a commit to minify `dist/main.js`. Should I just hold off until we figure out the direction we're going to go in?

joneshf-cn:
Nah, push it up!

joneshf-cn (approved):

PR #47, opened 2019-06-19
Add a `make test-watch` task

citizengabe:
## What it does

This adds a `make` task to watch `src/` and `test/` files and run `make test` after any changes.

## Review asks

It's not very efficient. I haven't looked closely, but I noticed that `make test` runs `psa` twice.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,4 +149,4 @@ watch: $(BOWER_COMPONENTS) $(NODE_MODULES)
 
 .PHONY: test-watch
 test-watch: $(BOWER_COMPONENTS) $(NODE_MODULES)
-	npx watch-exec --command 'make test' --watch $(TESTS) --watch $(SRC)
+	npx watch-exec --command 'make test' --watch $(TEST) --watch $(SRC)
--- a/Makefile
+++ b/Makefile
@@ -146,3 +146,7 @@ test: dist/main.js $(BUILD)/test.out
 .PHONY: watch
 watch: $(BOWER_COMPONENTS) $(NODE_MODULES)
 	npx watch-exec --command 'make dist/main.js' --watch $(EXAMPLE) --watch $(SRC)
+
+.PHONY: test-watch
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -146,3 +146,7 @@ test: dist/main.js $(BUILD)/test.out
 .PHONY: watch
 watch: $(BOWER_COMPONENTS) $(NODE_MODULES)
 	npx watch-exec --command 'make dist/main.js' --watch $(EXAMPLE) --watch $(SRC)
+
+.PHONY: test-watch
+test-watch: $(BOWER_COMPONENTS) $(NODE_MODULES)
+	npx watch-exec --command 'make test' --watch $(TESTS) --watch $(SRC)
```
joneshf-cn:
```suggestion
	npx watch-exec --command 'make test' --watch $(TEST) --watch $(SRC)
```
I think this should be `$(TEST)` instead of `$(TESTS)`–to parallel `$(SRC)`. I think if we state the individual files with `$(TESTS)`, it passes the whole `$(TESTS)` string as an argument, but the flag will only pick up the first file in the `$(TESTS)` string. It logs out something like:
```
[watch-exec] watching: test/Test/Lynx/Form.purs, src
```
It looks like it doesn't watch any of the other files in the `test` directory.

If we use `$(TEST)` there's only one argument to pass to it and it will watch all files in the `$(TEST)` directory. It logs out something like:
```
[watch-exec] watching: test, src
```
Which watches all files in `test`.

joneshf-cn (changes requested):
Thanks for this! I think there's one thing to change with the command. But, let me know if I'm missing anything.

> It's not very efficient. I haven't looked closely, but I noticed that `make test` runs `psa` twice.

Yeah, I think the dependencies aren't quite set up right. I think we depend on the src being fully built and then build the tests. I think we did this so CI could use `make test` and build all the targets. Probably better to have a CI target that does all that separately from the `test` target.

joneshf-cn (approved):

PR #48, opened 2019-06-20
Fix cache keys

joneshf-cn:
## What it does

We had a mismatch between the keys we save and find. It would be nice if this couldn't happen, but it doesn't look possible.

We set the keys to be accurate, for now.


citizengabe (approved):
LGTM

PR #49, opened 2019-06-21
Generate unique identifiers for sequences

joneshf-cn:
## What it does

As mentioned in [Slack][], we were falling victim to the common vdom array bug. We follow the solution described in the [React docs][], and use a unique key for each element.

You can replicate by:
1. Go to the "MVP" page
2. Go to the "Creative" tab
3. Click "Add" twice to add two elements
4. Fill out the inputs with different values
5. Remove the first element
1. [ ] Should only render the second element

[react docs]: https://reactjs.org/docs/lists-and-keys.html#keys
[slack]: https://citizennet.slack.com/archives/CGPP2STC4/p1561141723444200


src/Lynx/Form.purs
```diff
@@ -49,38 +49,58 @@ type Section f
...
   = { name :: String
     , key :: String
     , template :: TemplateSection f
-    , values :: InputSource (Array (Section f))
+    , values :: InputSource (Array (SequenceSection f))
     }
 
-stamp :: forall a. Key -> Page a -> Page a
-stamp key page = page { tabs = map stampTab page.tabs }
+stamp :: forall a f. Monad f => f String -> Key -> Page a -> f (Page a)
```
davezuch, line 65: I think we should agree on consistent type variables to avoid confusion. Throughout this file we use `f` to mean either `Expr` or `ExprType`. It's carried over from when our expressions were typed and `f` was the Functor wrapping our types `Expr a` vs `Identity a`. I don't know if we want to keep using `f` or rename it to something else, I've noticed `a` used instead in discussions and elsewhere.

Also in `src/Lynx.purs` we use `m` for our Monad. I'm okay with changing things, but I'd like to keep things consistent if we do.

davezuch (approved):
These changes look good to me

PR #51, opened 2019-06-25
PS-3032 Put together a list view

joneshf-cn:
## What it does

This is the first go at having a list view. It follows the ideas laid out in our [spec][], but only takes a slice of the whole picture.

## Review asks

We're only looking to solve PS-3032 with this one. There's still a bunch of work for PS-3023, but is there anything we missed for PS-3032?

[spec]: https://www.notion.so/citizennet/List-View-Config-9a5c51094f9343ecb4a1615f869c29a5


src/Lynx/List.purs
```diff
@@ -0,0 +1,216 @@
...
+component =
+  Halogen.component
+    { eval
+    , initialState
+    , receiver
+    , render
+    }
+
+eval :: forall f. Query ~> f
+eval query = eval query
```
davezuch: How does this work?

src/Lynx/List.purs
```diff
@@ -0,0 +1,216 @@
...
+
+fromArgonaut ::
+  forall a.
+  Data.Argonaut.DecodeJson a =>
+  Data.Argonaut.EncodeJson a =>
+  Data.Codec.Argonaut.JsonCodec a
+fromArgonaut = Data.Codec.Argonaut.prismaticCodec from to Data.Codec.Argonaut.json
+  where
+  from :: Data.Argonaut.Json -> Maybe a
+  from json = case Data.Argonaut.decodeJson json of
```
davezuch:
```purescript
from = Data.Either.hush <<< Data.Argonaut.decodeJson
```
?

davezuch: Or do we prefer the explicit casing?

src/Lynx/List.purs
```diff
@@ -0,0 +1,216 @@
...
+  lookup :: Lynx.Expr.Key -> Maybe Lynx.Expr.ExprType
+  lookup key = Nothing
+
+-- | This is the main thing that decides how to de/serialize our values.
+-- |
+-- | Something we run into from time to time with separate de/encoders is them
+-- | becoming out of sync.
+-- | It's a _mostly_ caught by the test suite,
+-- | but we can also allow the de/encoders to be correct by construction.
+jsonView :: Data.Codec.Argonaut.JsonCodec JSONView
```
davezuch, line 151: This is really cool, I didn't know it was something Argonaut supported

src/Lynx/List.purs
```diff
@@ -0,0 +1,216 @@
...
+  Right state ->
+    Ocelot.Block.Table.table_
+      [ Ocelot.Block.Table.row_ (map column state.columns)
+      ]
+  where
+  column :: RenderColumn -> Halogen.HTML a f
+  column column' =
+    Ocelot.Block.Table.header_
+      [ Halogen.HTML.text (Lynx.Expr.print column'.name)
+      ]
```
davezuch, line 206: I would think PS-3032 also involves rendering rows for a collection of `Array (Page a)` here, no?

davezuch (commented):
This looks good so far, but I think part of PS-3032 is still missing, as I mentioned in a comment below.

src/Lynx/List.purs
joneshf-cn: `Query _` is uninhabited. If someone were to construct a `Query _` and then passed it to `eval`, they've broken PureScript. There's no other implementation here than calling `eval` with the `Query _` they created.

src/Lynx/List.purs
joneshf-cn: I prefer casing, but I'm down to switch.

src/Lynx/List.purs
joneshf-cn: Makes sense, where would this live? Is it part of the list `View` or is it part of the `RenderView`?

src/Lynx/List.purs
joneshf-cn: Let's just make `Query _` a synonym for `Const Void _`. Hopefully that requires less explanation in this interim stage before we get more behavior.

src/Lynx/List.purs
davezuch: Ah it's no big deal!

src/Lynx/List.purs
davezuch: So if I understand your setup, `View` and `RenderView` are analogous to `View Expr` and `View ExprType`, if we were to do things more similar to `Lynx.purs`, yeah? In that case we just need to add `value :: ExprType` (or `String`) to `RenderView`, and add an eva that goes from `View -> RenderView`.

joneshf-cn: `View` and `RenderView` would be more like `Page Expr` and
```PureScript
type State
  = { activeTab :: String
    , evaled :: Either EvalError (Page ExprType)
    , expr :: Page Expr
    , fragment :: Fragment
    , idGenerator :: Aff String
    , values :: Map Key ExprType
    }
```
respectively. In any case, sounds like it belongs in the `RenderView`.

src/Lynx/List.purs
```diff
@@ -67,14 +69,17 @@ type Column
...
     }
 
 type RenderColumn
   = { name :: Lynx.Expr.ExprType
+    , value :: Lynx.Expr.Key
     }
 
 type Input
   = { columns :: Array Column
+    , rows :: Array (Lynx.Form.Page Lynx.Expr.ExprType)
```
davezuch, line 82: Can we assume that we're getting `Page ExprType`? I think the `GET` endpoint will return `Page Expr`, since that's what the form will expect, and may even be what we'll expect here. I guess this is one of those things we can kick down the road, but I do think we'll end up evaluating each row before rendering it.

davezuch (approved):
Changes look good!

src/Lynx/List.purs
joneshf-cn: Hmm, I'm not sure. This is what I was thinking: we can only create orders when they have all the values, those values would have to be someone typing in strings that we're evaluating, so we'd be storing the values as `ExprType` not parsing them to `Expr`. Is that not how it's going to work?

davezuch: > we can only create orders when they have all the values

Only when we have `required :: Boolean true` on all fields. But Spencer doesn't want us to worry about validation for now, so we're saving everything. Also, if every field has a `default :: Just _`, then we have values anyway. 

> so we'd be storing the values as `ExprType` not parsing them to `Expr`. Is that not how it's going to work?

I think we should always store values as `ExprType`, even for `Page Expr`. I don't think a value of `InputSource Expr` ever makes sense, as users can't input expressions (maybe in the Form Builder form it'd make sense). 

But returning `Page Expr` instead of `Page ExprType` is necessary, at least for the form, because it needs to know what the expressions are in order to evaluate changes to the form. At the least we'd need to send both, but that seems redundant because we can get one from the other.

For the list view it's less clear that `Page Expr` is necessary, but I argue that it is, because evaluation isn't pure, there are HTTP lookups that can happen, or date validations that might happen, which means the result of evaluating now might be different from the result of evaluating when the Order was saved. I guess which result is more correct is debatable, but I think the newer one.

joneshf-cn: I see. It's important to note that whole thing isn't accurate anyway. We should be working with an Order, not a `Page _` (whatever that is in the OMS parlance). I think no matter what we put in here, it'll be an approximation of what the real thing is until we get to that.

PR #52, opened 2019-06-27
Prevent running purty as root

davezuch:
## What it does

I don't know how but I ended up with all my `.purs` files being owned by `root` and had to nuke everything. This should hopefully prevent me from doing a dumb thing in the future.

## Review asks

This seem sensible?


Makefile
```diff
@@ -112,9 +115,13 @@ $(BUILD)/test.out: $(BUILD)/test.js | $(BUILD)
 	mv $@.tmp $@ # Move the output where it belongs.
 
 $(FORMAT)/%.purs.formatted: %.purs $(NODE_MODULES)
+ifeq '$(USER)' 'root'
+	@echo "$(RED)purty should not be ran as root, skipping"
```
citizengabe: "should not be _run_ as root", but otherwise 👍

citizengabe (changes requested):

Makefile
davezuch: 🙃

citizengabe (approved):

PR #53, opened 2019-07-02
PS-3034 - Adjustable Column Widths

davezuch:
## What it does

Introduces way to specify column widths, and adds appropriate rendering. Updates example.

## How should this be manually tested?

`make deploy` then navigate to `/#/list`.


src/Lynx/List.purs
```diff
@@ -61,6 +65,7 @@ type View
 type Column
   = { name :: Lynx.Expr.Expr
     , value :: Lynx.Expr.Key
+    , width :: Maybe Size
```
joneshf-cn: Can we convert from `Maybe Size` to more domain-specific terminology? I think we decided on: auto, small, medium, and large as the different widths. It would be nice to speak the same language as the spec when we're working within the domain. It's fine for the JSON side of things to remain with primitives like `Maybe _`, though.

src/Lynx/List.purs
```diff
@@ -75,8 +80,39 @@ type RenderView
...
+  | Large
+
+derive instance eqSize :: Eq Size
+
+derive instance genericSize :: Generic Size _
+
+instance showSize :: Show Size where
+  show = genericShow
+
+instance encodeJsonSize :: Data.Argonaut.EncodeJson Size where
```
joneshf-cn: Can we drop the `DecodeJson`/`EncodeJson` instances in favor of using codecs? The intent behind using the codecs is that we then don't need to write a roundtrip test for the JSON side of things. We should have enough confidence that what we're encoding and decoding works with each other.

If we go with the `DecodeJson`/`EncodeJson` approach, we have to validate that those line up with a roundtrip test like we've been doing for all the JSON stuff in `Lynx.Expr` and `Lynx.Form`. Which  means that we'll have to add a test suite for `Lynx.List` round tripping.

joneshf-cn (commented):
Looks good!

I have a couple of comments, but looks fine otherwise.

joneshf-cn (approved):

PR #54, opened 2019-07-11
PS-3041 Add form buttons

davezuch:
## What it does

Adds support for a "cancel" and "save" button that raise a message to the parent component. This is necessary for PS-3041, as it should allow the parent component to then save a form or navigate away.

## Review asks

Is this how we want to handle these events? We could do something more fancy like telling Lynx how to save or where to route upon cancel, but I don't feel that's necessary. I think in this case it's better to keep control of that in the parent.


joneshf-cn (approved):
Looks good!

> Is this how we want to handle these events? We could do something more fancy like telling Lynx how to save or where to route upon cancel, but I don't feel that's necessary. I think in this case it's better to keep control of that in the parent.

Yeah, makes sense.

davezuch:
Thanks for reviewing!

